PR #65, opened 2017-08-04
Fix legacy accessors

jfparadis:
In this request we:
- Fix two infinite loops
- Remove a redundant seal/freeze
- Set the stage for more sanitization

### DIFF
--- a/shim/src/sanitize.js
+++ b/shim/src/sanitize.js
@@ -1,49 +1,10 @@
+import repairAccessors from "./repair/accessors";
+
 // locking down the environment
 export function sanitize(sandbox) {
-    const { confinedWindow: { Object: o } } = sandbox;
-    try {
-        // Fixing properties of Object to comply with strict mode
-        // and ES2016 semantics, we do this by redefining them while in 'use strict'
-        // https://tc39.github.io/ecma262/#sec-object.prototype.__defineGetter__
-        if (o === undefined) {
-            return;
-        }
-        o.defineProperty(o.prototype, '__defineGetter__', {
-            value: function (key, fn) {
-                return o.defineProperty(this, key, {
-                    get: fn
-                });
-            }
-        });
-        o.defineProperty(o.prototype, '__defineSetter__', {
-            value: function (key, fn) {
-                return o.defineProperty(this, key, {
-                    set: fn
-                });
-            }
-        });
-        o.defineProperty(o.prototype, '__lookupGetter__', {
-            value: function (key) {
-                var d, p = this;
-                while (p && (d = o.getOwnPropertyDescriptor(p, key)) === undefined) {
-                    p = o.getPrototypeOf(this);
-                }
-                return d ? d.get : undefined;
-            }
-        });
-        o.defineProperty(o.prototype, '__lookupSetter__', {
-            value: function (key) {
-                var d, p = this;
-                while (p && (d = o.getOwnPropertyDescriptor(p, key)) === undefined) {
-                    p = o.getPrototypeOf(this);
-                }
-                return d ? d.set : undefined;
-            }
-        });
-        // Immutable Prototype Exotic Objects
-        // https://github.com/tc39/ecma262/issues/272
-        o.seal(o.prototype);
-    } catch (ignore) {
-        // Ignored
-    }
+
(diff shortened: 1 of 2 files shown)


shim/src/sanitize.js
```diff
@@ -1,49 +1,7 @@
...
-                return o.defineProperty(this, key, {
-                    set: fn
-                });
-            }
-        });
-        o.defineProperty(o.prototype, '__lookupGetter__', {
-            value: function (key) {
-                var d, p = this;
-                while (p && (d = o.getOwnPropertyDescriptor(p, key)) === undefined) {
-                    p = o.getPrototypeOf(this);
```
jfparadis: This line was creating an infinite loop. The correct code should be:
```
p = o.getPrototypeOf(p);
```

shim/src/sanitize.js
```diff
@@ -1,49 +1,7 @@
...
-                }
-                return d ? d.set : undefined;
-            }
-        });
-        // Immutable Prototype Exotic Objects
-        // https://github.com/tc39/ecma262/issues/272
-        o.seal(o.prototype);
```
jfparadis: Immutability of Object will be handled elsewhere, together with other objects accessible by grammar.

shim/src/sanitize/accessors.js
```diff
@@ -0,0 +1,62 @@
+// Adapted from SES/Caja - Copyright (C) 2011 Google Inc.
+// https://github.com/google/caja/blob/master/src/com/google/caja/ses/startSES.js
+// https://github.com/google/caja/blob/master/src/com/google/caja/ses/repairES5.js
```
jfparadis: We should give credit to the original implementation, even if this is small subset and modernized version of the code the code. Doing so will also help with continuity.

shim/src/sanitize.js
```diff
@@ -1,49 +1,7 @@
...
-                    p = o.getPrototypeOf(this);
-                }
-                return d ? d.get : undefined;
-            }
-        });
-        o.defineProperty(o.prototype, '__lookupSetter__', {
-            value: function (key) {
-                var d, p = this;
-                while (p && (d = o.getOwnPropertyDescriptor(p, key)) === undefined) {
-                    p = o.getPrototypeOf(this);
```
jfparadis: Infinite loop.

shim/src/sanitize/accessors.js
```diff
@@ -0,0 +1,62 @@
...
+
+import { getPrototypeOf, defineProperty, getOwnPropertyDescriptor } from '../commons';
+
+// Fix legacy accessors to comply with strict mode and ES2016 semantics,
+// we need to redefine them while in strict mode.
+// https://tc39.github.io/ecma262/#sec-object.prototype.__defineGetter__
+
+export default function sanitizeAccessors(sandbox) {
+
+  const { confinedWindow: { Object: { prototype: proto } } } = sandbox;
```
caridy: if the role of this function is to sanitize accessors, then pass the `Object.prototype` to it instead of passing sandbox, which is our own invention.

shim/src/sanitize.js
```diff
@@ -1,49 +1,7 @@
+import sanitizeAccessors from "./sanitize/accessors";
```
caridy: I don't like that now we have `sanitizy.js` and `sanitize/*.js`, maybe this folder should not be called `sanitize` but something else were we can drop more similar files in the future.

shim/src/sanitize/accessors.js
```diff
@@ -0,0 +1,62 @@
+// Adapted from SES/Caja - Copyright (C) 2011 Google Inc.
```
caridy: noticed that most files are using 4 spaces, and this one is using 2, maybe we should just add a editorconfig file to the project.

shim/src/sanitize/accessors.js
jfparadis: ok

shim/src/sanitize/accessors.js
jfparadis: or a linter?

shim/src/sanitize.js
jfparadis: I do like this approach, similar to Java and other languages. But I'll go with the flow.

PR #66, opened 2017-08-04
Add sanitize functions

jfparadis:
Gradually adapting and merging Mark's code:
https://github.com/tc39/proposal-realms/issues/54

### DIFF
--- a/shim/src/evaluators.js
+++ b/shim/src/evaluators.js
@@ -9,27 +9,12 @@ function getEvalEvaluator(sandbox) {
             return evaluate(sourceText, sandbox);
         }
     };
-    setPrototypeOf(o.eval, sandbox.Function);
+    // TODO
+    // setPrototypeOf(o.eval, sandbox.Function);
     o.eval.toString = () => 'function eval() { [shim code] }';
     return o.eval;
 }
 
-function getFunctionEvaluator(sandbox) {
-    const { confinedWindow } = sandbox;
-    const f = function Function(...args) {
-        // console.log(`Shim-Evaluation: Function("${args.join('", "')}")`);
-        const sourceText = args.pop();
-        const fnArgs = args.join(', ');
-        return evaluate(`(function anonymous(${fnArgs}){\n${sourceText}\n}).bind(this)`, sandbox);
-    }
-    f.prototype = confinedWindow.Function.prototype;
-    setPrototypeOf(f, f.prototype);
-    f.prototype.constructor = f;
-    f.toString = () => 'function Function() { [shim code] }';
-    return f;
-}
-
 export function getEvaluators(sandbox) {
-    sandbox.Function = getFunctionEvaluator(sandbox);
     sandbox.eval = getEvalEvaluator(sandbox);
 }
--- a/shim/src/repair/functionConstructors.js
+++ b/shim/src/repair/functionConstructors.js
@@ -1,0 +1,68 @@
+// Adapted from SES/Caja - Copyright (C) 2011 Google Inc.
+// https://github.com/google/caja/blob/master/src/com/google/caja/ses/startSES.js
+// https://github.com/google/caja/blob/master/src/com/google/caja/ses/repairES5.js
+
+import verifyStrictFunctionBody from '../evaluate/verifyStrictFunctionBody';
+import { getPrototypeOf, setPrototypeOf, defineProperty } from '../commons';
+
+// This block replaces the original Function, %GeneratorFunction% and
+// %AsycnGeneratorFunction%, with safe implementations that preserve SES
+// confinement. After this block is done, the originals should no longer be
+// reachable.
+
+export default function repairFunctionConstructors(confinedWindow) {
+
+    const { eval: secureEval } = confinedWindow;
+
+    /**
+     * A safe form of the {@code Function} constructor, which
+     * constructs strict functions that can only refer freely to the
+     * {@code sharedImports}.
+     *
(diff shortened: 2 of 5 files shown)


shim/src/evaluators.js
```diff
@@ -9,27 +9,12 @@ function getEvalEvaluator(sandbox) {
...
-    }
-    f.prototype = confinedWindow.Function.prototype;
-    setPrototypeOf(f, f.prototype);
-    f.prototype.constructor = f;
-    f.toString = () => 'function Function() { [shim code] }';
-    return f;
-}
-
 export function getEvaluators(sandbox) {
-    sandbox.Function = getFunctionEvaluator(sandbox);
```
jfparadis: Now handled by "repairFunctionConstructors"

shim/src/repair/functionConstructors.js
```diff
@@ -0,0 +1,68 @@
...
+
+import verifyStrictFunctionBody from '../evaluate/verifyStrictFunctionBody';
+import { getPrototypeOf, setPrototypeOf, defineProperty } from '../commons';
+
+// This block replaces the original Function, %GeneratorFunction% and
+// %AsycnGeneratorFunction%, with safe implementations that preserve SES
+// confinement. After this block is done, the originals should no longer be
+// reachable.
+
+export default function repairFunctionConstructors(confinedGlobal) {
```
jfparadis: I'm toying with "confinedGlobal" instead of "confinedWindow". If you agree, I will make replacements everywhere.

shim/src/repair/objectAccessors.js
```diff
@@ -8,7 +8,9 @@ import { getPrototypeOf, defineProperty, getOwnPropertyDescriptor } from '../com
 // we need to redefine them while in strict mode.
 // https://tc39.github.io/ecma262/#sec-object.prototype.__defineGetter__
 
-export default function repairAccessors(objProto) {
+export default function repairObjectAccessors(confinedGlobal) {
+
+    const { Object: { prototype: objProto } } = confinedGlobal;
```
jfparadis: I'll have all repair functions use the same arguments, the global, but be more specific about what they do.

shim/src/repair/functionConstructors.js
```diff
@@ -0,0 +1,68 @@
...
+    // Our process to secure constructors:
+    // 1. Obtain the original prototype of an instance from the sandbox
+    // 2. Create a secure constructor
+    // 3. Replace its prototype property with the original prototype
+    // 4*. Replace its internal [[Prototype]] property with a secure function (except for Function)
+    // 5. Replace the original prototype's constructor with the secure constructor
+    // 6*. Export the secure function to the sandbox (only for Function)
+
+    const FuncProto = getPrototypeOf(secureEval('(function () {})'));
+    const secureFunction = secureFunctionFactory('function');
```
caridy: I think you mean upper-case `'Function'`

shim/src/repair/functionConstructors.js
```diff
@@ -0,0 +1,68 @@
...
+     * A safe form of the {@code Function} constructor, which
+     * constructs strict functions that can only refer freely to the
+     * {@code sharedImports}.
+     *
+     * The returned function is strict whether or not it declares
+     * itself to be.
+     */
+    function secureFunctionFactory(funcDeclaration) {
+        return (function SecureFunction(...params) {
+            var body = verifyStrictFunctionBody(params.pop() || '');
```
caridy: I don't fully understand this code...  let's chat about it

shim/src/evaluators.js
```diff
@@ -9,27 +9,12 @@ function getEvalEvaluator(sandbox) {
             return evaluate(sourceText, sandbox);
         }
     };
-    setPrototypeOf(o.eval, sandbox.Function);
+    // TODO
+    // setPrototypeOf(o.eval, sandbox.Function);
```
caridy: where are you doing this now?

shim/src/repair/functionConstructors.js
jfparadis: The code is correct. I could have used the alternate syntax:
```
Reflect.getPrototypeOf(eval('(function () {})')) === global.Function.prototype
// true
```

But using `Function` would not have been right. `Function` and `Function.prototype` are different. However, `Function` and `Function.constructor` are the same.

To avoid that trap and to ease maintenance, I wanted the code to be symmetric, since the "repair" process is complex but always the same. I repeatedly use the same technique for Function, FunctionGenerator and AsyncFunction.  

The only difference with Mark's original [code](https://github.com/tc39/proposal-realms/issues/54) is to support a given `sandbox`:
```
const UnsafeFunc = Function;
```
That code was not written to use a provided context, such as "sandbox", but was using the `Function` as visible in the current execution context.

Idem with FunctionGenerator which was also using the current execution context:
```
const Generator = Reflect.getPrototypeOf(function*(){});
```

shim/src/repair/functionConstructors.js
jfparadis: See `makeFuncConstructor`  in Mark's code:
[code](https://github.com/tc39/proposal-realms/issues/54):

shim/src/evaluators.js
jfparadis: Because it's a TODO!

shim/src/repair/functionConstructors.js
caridy: oh, I understand now. This is reused for function and function*, and any other form of function declaration. got it.

PR #75, opened 2018-02-12
Improve stdlib

jfparadis:
Added missing object, added comments to source.

### DIFF
--- a/shim/src/stdlib.js
+++ b/shim/src/stdlib.js
@@ -2,18 +2,35 @@ import { getIntrinsics } from "./intrinsics";
 
 export function getStdLib(sandbox) {
     const intrinsics = getIntrinsics(sandbox);
+
     return {
+        // *** 18.1 Value Properties of the Global Object
+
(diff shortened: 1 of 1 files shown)


shim/src/stdlib.js
```diff
@@ -1,55 +1,77 @@
...
-        WeakMap: { value: intrinsics.WeakMap },
-        WeakSet: { value: intrinsics.WeakSet },
-
-        // TODO: Annex B
-        // TODO: other special cases
-    };
+
+  const intrinsics = getIntrinsics(sandbox);
+
+  return {
```
caridy: wrong alignment causing the diff to be bigger that it is. The entire project is using 4 spaces, but this file is using 2.

jfparadis:
Do you want to keep that 4x? It's kindof odd nowadays.

caridy:
unless that there is a reason for changing, we should stick to what we have, specially because it preserves the history of the file more accurate. w=1 trick does not allow you to comment on the line anyway.

PR #79, opened 2018-02-18
Completed intrinsics part 2

jfparadis:


shim/src/intrinsics.js
```diff
@@ -285,7 +281,7 @@ export function getIntrinsics(sandbox) {
         // %unescape%
         unescape: global.unescape,
 
-        // TODO: other special cases
+        // TODOther special cases
```
chicoxyzzy: Was this caused by some automatic text text replacement?

jfparadis: Thanks. Fixed. Appreciated!

PR #101, opened 2018-05-24
Initial unit tests

jfparadis:
Looking at a way to structure this. In particular, we need to check that the building process doesn't distortions on the expected behaviour.


ljharb:
The approach LGTM

PR #111, opened 2018-06-18
rename/refactor shim components

warner:
This rearranges a lot of the shim, and adds some questions and TODOs. It's involved enough that we should probably walk through it and discuss each section, rather than simply landing it directly.


warner:
This builds on top of #110, so we should probably land that before looking too closely at this one.

PR #117, opened 2018-06-18
Jparadis/master/update realm construction

jfparadis:


warner:
looks good to me

shim/src/realm.js
```diff
@@ -72,6 +71,16 @@ function buildChildRealm(BaseRealm) {
     evaluate(...args) {
       return doAndWrapError(() => descs.evaluate.value.apply(this, args));
     }
+    static makeRootRealm() {
```
caridy: Late to the party here, but I wonder why are we adding new APIs to the shim that are not part of the realm spec, instead of just adding them to an abstraction layer on top of the shim?

shim/src/realm.js
```diff
@@ -165,15 +190,6 @@ export default class Realm {
     };
     Realm2RealmRec.set(O, realmRec);
 
-    const init = O.init;
-    if (!IsCallable(init)) throw new TypeError();
-    init.call(O);
```
caridy: I still don't fully understand how are we removing `init`, and how do the user have control of the globals added to global during initialization.

It seems that the shim is moving faster than the spec, which is fine, but maybe it should not be part of this repo then. My main concern is that folks who come here to look at the spec, and try out the polyfill, will not realize that the polyfill is not following the spec.

PR #120, opened 2018-06-19
shim: avoid making cycles in repairFunction

warner:
Use `Object.*` instead of `Reflect.*` to catch more errors like this.


warner:
Looks like a temporary failure.. I don't have the authority to push the "Restart Build" button, but @jfparadis maybe you do?

PR #122, opened 2018-06-19
Remove unsafeRec from realmRec

jfparadis:
Consistently use deconstruction
Remove safeEvaluators, define safeEval and safeFunction on realmRec
Collapse sanitize into context


warner:
looks good to me

PR #125, opened 2018-06-21
Rework 

jfparadis:


warner:
we reviewed this in person, looks good, we have some changes to apply on top

PR #126, opened 2018-06-22
more todo notes, forEach fixes

warner:
a few more notes from today's review

### DIFF
--- a/shim/src/scan.js
+++ b/shim/src/scan.js
@@ -73,9 +73,11 @@ export function walkObjects(start, visitor) {
 
     enqueue(getPrototypeOf(obj), obj, '__proto__');
     const descs = getOwnPropertyDescriptors(obj);
-    ownKeys(descs).forEach(name => {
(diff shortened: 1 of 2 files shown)


shim/src/scan.js
```diff
@@ -73,9 +73,11 @@ export function walkObjects(start, visitor) {
 
     enqueue(getPrototypeOf(obj), obj, '__proto__');
     const descs = getOwnPropertyDescriptors(obj);
-    ownKeys(descs).forEach(name => {
+    for (let name of ownKeys(descs)) {
+      // todo: all iteration needs uncurried forEach protection
```
ljharb: using `for..of` means you also need to protect against `Symbol.iterator` being poisoned, no?

warner: Probably ,yeah.  I think we decided that we should just capture an uncurried-this version of `Array.forEach` in `commons.js` and use that everywhere we do iteration.

PR #128, opened 2018-06-25
Fix Function with no arguments

jfparadis:
Fix for:
```js
(new Function()).toString();

function() {
undefined
}
```

### DIFF
--- a/shim/src/evaluators.js
+++ b/shim/src/evaluators.js
@@ -180,7 +180,7 @@ export function createFunctionEvaluator(unsafeRec, safeEval) {
   const { unsafeFunction, unsafeGlobal } = unsafeRec;
 
   const safeFunction = function Function(...params) {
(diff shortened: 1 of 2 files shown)


jfparadis:
@warner Could you review?

warner:
Looks good to me. Maybe have the test case make sure the function can be invoked too? I guess it should return undefined.

warner:
maybe add:

```
diff --git a/shim/test/realm/function-eval.js b/shim/test/realm/function-eval.js
index 891fe3d..3d649b2 100644
--- a/shim/test/realm/function-eval.js
+++ b/shim/test/realm/function-eval.js
@@ -6,6 +6,7 @@ test('function-no-body', t => {
   const f1 = new r.global.Function();
   const src = f1.toString();
   t.notOk(src.includes('undefined'));
+  t.equal(f1(), undefined);
   t.end();
 });
```

PR #130, opened 2018-06-27
Fix a typo: inherited -> inherit

koba04:


shim/src/realm.js
```diff
@@ -164,7 +164,7 @@ class BaseRealm {
       // note this would leak the parent TypeError, from which the child can
       // access .prototype and the parent's intrinsics, except that the Realm
       // facade catches all errors and translates them into local Error types
-      throw new TypeError('Realm only supports undefined or "inherited" intrinsics.');
+      throw new TypeError('Realm only supports undefined or "inherit" intrinsics.');
```
ljharb: this seems correct to me as-is, since both are adjectives

koba04: I guess this value points the string value.

https://github.com/tc39/proposal-realms/blob/master/shim/src/realm.js#L128

ljharb: ah, then I’d backtick inherit, to show it’s not part of the English prose :-)

caridy: that term was always confusing, and we should probably either change it, or change the way you signal the inheritance mechanism for intrinsics. But for now, I think we can stick to `inherit`.

warner: FWIW one of the PRs @erights and I will be filing today removes that option entirely in favor of direct `Realm.makeRootRealm()` and `Realm.makeCompartment()` calls. Modifying the new Realm is done with injected shims, since we found it's pretty hard to safely modify the intrinsics in-place or wrap them between realms.

jfparadis:
API changed. Good discussion still.

PR #131, opened 2018-07-02
commons: use faster implementation of uncurryThis()

warner:
This version should be faster on most platforms.

### DIFF
--- a/shim/src/commons.js
+++ b/shim/src/commons.js
@@ -26,8 +26,13 @@ export const {
 
 // See http://wiki.ecmascript.org/doku.php?id=conventions:safe_meta_programming
 // which only lives at http://web.archive.org/web/20160805225710/http://wiki.ecmascript.org/doku.php?id=conventions:safe_meta_programming
-const bind = Function.prototype.bind;
-const uncurryThis = bind.bind(bind.call);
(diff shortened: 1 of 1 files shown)


shim/src/commons.js
```diff
@@ -26,8 +26,13 @@ export const {
...
 // which only lives at http://web.archive.org/web/20160805225710/http://wiki.ecmascript.org/doku.php?id=conventions:safe_meta_programming
-const bind = Function.prototype.bind;
-const uncurryThis = bind.bind(bind.call);
+// (the native call is about 10x faster on FF than chrome)
+// this version of uncurryThis is about 100x slower on FF, equal on chrome, 2x slower on Safari
+// const bind = Function.prototype.bind;
+// const uncurryThis = bind.bind(bind.call);
+
+// this version is about 10x slower on FF, equal on chrome, 2x slower on Safari
+const uncurryThis = fn => (thisArg, ...args) => apply(fn, thisArg, args);
```
caridy: I'm fine with this change, but keep in mind that the slowness might be coming from the rest argument, maybe just doing `Array.prototype.slice.call(arguments, 2)` might be faster across the board.

PR #132, opened 2018-07-02
evaluators: protect 'for of' against poisoning by using arrayForEach

warner:


jfparadis:
There is a little more than uncurrying, we are also converting `getOptimizableGlobals` to use it.

PR #147, opened 2018-07-05
Add code coverage

jfparadis:
use `npm run shim:coverage`


ljharb:
I'd enforce whatever percentages you have now :-) then ratchet that up over time as coverage increases

PR #149, opened 2018-07-06
Simplify Accessors (DRY)

jfparadis:
@warner Just some cleanup for maintainability

### DIFF
--- a/shim/src/accessors.js
+++ b/shim/src/accessors.js
@@ -4,35 +4,56 @@
 
 /**
  * Replace the legacy accessors of Object to comply with strict mode
- * and ES2016 semantics, we do this by redefining them while in 'use strict'
- * https://tc39.github.io/ecma262/#sec-object.prototype.__defineGetter__
-
-We need this repair, but would it be included when the real realms is
-integrated into the language. If not, what are we getting here?
-
-Also note that this changes the primal versions.
-
-On some platforms, the implementation of these functions act as if they are
-in sloppy mode: if they're invoked badly, they will expose the global object,
-so we need to repair these for security. Thus it is our responsibility to fix
-this, and we need to include repairAccessors. E.g. Chrome in 2016.
-
-todo: It would be better to detect and only repair the functions that have
-the bug.
+ * and ES2016 semantics, we do this by redefining them while in 'use strict'.
+ *
+ * todo: list the issues resolved
+ *
+ * This function can be used in two ways: (1) invoked directly to fix the primal
+ * realm's Object.prototype, and (2) converted to a string to be executed
+ * inside each new RootRealm to fix their Object.prototypes. Evaluation requires
+ * the function to have no dependencies, so don't import anything from the outside.
  */
 
-// todo: this file should be moved out to a separate repo and npm module
-
-// We use this function in two ways. We use it directly to fix the primal
-// realm's Object.prototype, and we convert it into a string to be executed
-// inside each new RootRealm to fix their Object.prototypes too. So don't
-// import anything from the outside.
+// todo: this file should be moved out to a separate repo and npm module.
 
 // todo: This function is serialized and evaluated outside of the primal
 // realms and it currently can't contain code coverage metrics.
 /* istanbul ignore file */
 export function repairAccessors() {
-  const { getPrototypeOf, defineProperties, defineProperty, getOwnPropertyDescriptor } = Object;
+  const {
+    getPrototypeOf,
+    defineProperty,
+    getOwnPropertyDescriptor,
+    prototype: objectPrototype
+  } = Object;
+
+  // On some platforms, the implementation of these functions act as if they are
+  // in sloppy mode: if they're invoked badly, they will expose the global object,
+  // so we need to repair these for security. Thus it is our responsibility to fix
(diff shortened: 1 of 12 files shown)


shim/src/accessors.js
```diff
@@ -43,47 +43,29 @@ export function repairAccessors() {
...
+    let base = this;
+    let desc;
+    while (base && !(desc = getOwnPropertyDescriptor(base, prop))) {
+      base = getPrototypeOf(base);
     }
+    return desc && desc.get;
+  }
+
+  defineProperties(Object.prototype, {
+    __defineGetter__: { value: defineAccessor },
```
ljharb: doesn't this mean that `__defineGetter__ === __defineSetter__`, which isn't supposed to be true?

jfparadis: Good catch. Let me see.

jfparadis:
Perfomance comparison:
https://jsperf.com/dry-repair-accessors

Difference at most of 5% of 50M ops/sec (5% of 0.000,02ms) on FF

shim/src/accessors.js
```diff
@@ -43,26 +43,20 @@ export function repairAccessors() {
     return `${prop}`;
   }
 
-  defineProperties(Object.prototype, {
-    __defineGetter__: {
+  ['__defineGetter__', '__defineSetter__'].forEach(name =>
+    defineProperty(prototype, name, {
       value(prop, func) {
         return defineProperty(this, prop, {
           get: func,
```
erights: For the `__defineSetter__` case, this needs to be `set:`, not `get:`

shim/src/accessors.js
```diff
@@ -72,19 +66,8 @@ export function repairAccessors() {
         }
         return desc && desc.get;
```
erights: For the `__lookupSetter__` case, this needs to be `desc.set`, not `desc.get`.

erights:
The fact that the tests pass anyway indicates what tests we're missing

jfparadis:
Sorry for the breaking submission: I was focusing on code complexity metrics. 
I hear you regarding tests. Stay tuned, I will add them with this PR.

shim/src/accessors.js
```diff
@@ -4,35 +4,43 @@
...
+
+  // On some platforms, the implementation of these functions act as if they are
+  // in sloppy mode: if they're invoked badly, they will expose the global object,
+  // so we need to repair these for security. Thus it is our responsibility to fix
+  // this, and we need to include repairAccessors. E.g. Chrome in 2016.
+
+  function makeDefineAccessor(method, accessor) {
+    defineProperty(prototype, method, {
+      value(prop, func) {
+        defineProperty(this, prop, {
```
jfparadis: According to specs, we should not return here.

jfparadis:
@erights Tests are in, simply a subset of test262

shim/test/test262/annexB/built-ins/Object/prototype/__defineGetter__/define-new.js
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
```
jfparadis: I'm still studying the best pattern to bring test262 in. This is useful as is but most likely to evolve.

shim/src/accessors.js
jfparadis: It's supposed to throw (and I added the test):
```
esid: sec-additional-properties-of-the-object.prototype-object
description: Behavior when property exists and is not configurable
info: |
    [...]
    5. Perform ? DefinePropertyOrThrow(O, key, desc).
```

shim/src/accessors.js
```diff
@@ -4,35 +4,43 @@
...
-// import anything from the outside.
-
 // todo: This function is serialized and evaluated outside of the primal
 // realms and it currently can't contain code coverage metrics.
 /* istanbul ignore file */
 export function repairAccessors() {
-  const { getPrototypeOf, defineProperties, defineProperty, getOwnPropertyDescriptor } = Object;
+  const { getPrototypeOf, defineProperty, getOwnPropertyDescriptor, prototype } = Object;
+
+  // On some platforms, the implementation of these functions act as if they are
```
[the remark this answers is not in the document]
jfparadis: Added comment. FYII: I could not find the detection code in caja or ses.

shim/src/accessors.js
```diff
@@ -4,35 +4,43 @@
...
-// realm's Object.prototype, and we convert it into a string to be executed
-// inside each new RootRealm to fix their Object.prototypes too. So don't
-// import anything from the outside.
-
 // todo: This function is serialized and evaluated outside of the primal
 // realms and it currently can't contain code coverage metrics.
 /* istanbul ignore file */
 export function repairAccessors() {
-  const { getPrototypeOf, defineProperties, defineProperty, getOwnPropertyDescriptor } = Object;
+  const { getPrototypeOf, defineProperty, getOwnPropertyDescriptor, prototype } = Object;
```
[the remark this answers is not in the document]
jfparadis: Added handling for "false". 
Copied a test262 to check for throw.
Renamed constant.

PR #150, opened 2018-07-07
Simplify optimizer

jfparadis:
Aka optimize the optimizer...

### DIFF
--- a/shim/src/evaluators.js
+++ b/shim/src/evaluators.js
@@ -4,15 +4,14 @@
 import {
   apply,
   arrayJoin,
+  arrayFilter,
   arrayPop,
-  arrayPush,
   create,
   defineProperty,
   getOwnPropertyDescriptors,
   getOwnPropertyNames,
   getPrototypeOf,
   regexpMatch,
-  arrayForEach,
   setPrototypeOf,
   stringIncludes
 } from './commons';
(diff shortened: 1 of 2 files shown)


shim/src/evaluators.js
```diff
@@ -4,70 +4,67 @@
...
-    // Check for accessor properties: we don't want to optimize these,
-    // they're obviously non-constant. Setter-only accessors will still have
-    // a 'get' property, but it will be 'undefined', so we only have to test
-    // for 'get', not 'set'
-    if ('get' in desc) return;
-    if ('set' in desc) return;
+  const constants = arrayFilter(getOwnPropertyNames(descs), name => {
+    //
+    // getOwnPropertyNames does ignore Symbols so we don't need this extra check:
+    // typeof name === 'string' &&
```
jfparadis: Important comment above. We need to add tests to guard the behavior.

shim/src/evaluators.js
```diff
@@ -4,70 +4,67 @@
...
-    // protect against post-initialization corruption of primal realm Array
-    arrayPush(constants, name);
+    return false;
   });
+
   return constants;
 }
 
 function buildOptimizer(constants) {
+  if (constants.length === 0) return '';
```
jfparadis: Just to avoid "const {} = arguments[0]";

shim/src/evaluators.js
```diff
@@ -4,70 +4,67 @@
...
-    if ('set' in desc) return;
+  const constants = arrayFilter(getOwnPropertyNames(descs), name => {
+    //
+    // getOwnPropertyNames does ignore Symbols so we don't need this extra check:
+    // typeof name === 'string' &&
+    if (regexpMatch(identifierPattern, name)) {
+      // Optimization: retrieve the descriptor only if we have a valid identifier.
+      const desc = descs[name];
+
+      return (
```
jfparadis: I'm using code complexity metrics and this whole function was flagged has having too many return statements. Looking into it, there were several opportunities for optimization.

warner: Huh.. if it's faster with the new control flow, that sounds like a good reason to use it, but the new one is less obvious to me: it's combining an if/then and a short-circuiting `&&`. Would it be equally as fast to replace the inner `return(x&&y&&z)` with an `if (x&&y&&z) return true`, and otherwise fall through to the final `return false` outside the `if (regexpMatch)` block?

jfparadis: Sorry, I am not following that.

shim/src/evaluators.js
```diff
@@ -4,70 +4,65 @@
...
 
 // todo: think about how this interacts with endowments, check for conflicts
 // between the names being optimized and the ones added by endowments
 
 function getOptimizableGlobals(safeGlobal) {
-  const constants = [];
   const descs = getOwnPropertyDescriptors(safeGlobal);
 
-  arrayForEach(getOwnPropertyNames(descs), name => {
+  const constants = arrayFilter(getOwnPropertyNames(descs), name => {
```
jfparadis: @warner Here is an update that might address what I think is yourc concern.

warner: yeah, that's marginally better.. I'd personally lean towards a list of `if x return false` clauses, but I'm happy with what you've got here too.

shim/src/evaluators.js
warner: I guess I was thinknig of something like this:
```
const constants = arrayFilter(getOwnPropertyNames(descs), name => {
  if (regexpMatch(identifierPattern, name)) {
    const desc = descs[name];
    if (desc.configurable === true &&
        desc.writable === true &&
        'value' in desc) {
      return true;
    }
  }
  return false;
});
```

shim/src/evaluators.js
```diff
@@ -4,70 +4,65 @@
...
   setPrototypeOf,
   stringIncludes
 } from './commons';
 import { ScopeHandler } from './scopeHandler';
 import { rejectImportExpressions } from './block-imports';
 import { assert, throwTantrum } from './utilities';
 
+// admit many (but not all) legal variable names: starts with letter/_/$,
+// continues with letter/digit/_/$ . It will reject many legal names that
+// involve unicode characters. We use 'apply' rather than /../.match() in
```
erights: Comment stale. Should be 

We use 'regexpMatch' rather than

shim/src/evaluators.js
```diff
@@ -4,70 +4,65 @@
...
-    if ('get' in desc) return;
-    if ('set' in desc) return;
-
-    // protect against post-initialization corruption of primal realm Array
-    arrayPush(constants, name);
+    return (
+      //
+      // getters will not have .writable, don't let the falsyness of
+      // 'undefined' trick us: test with === false, not ! . However descriptors
+      // inherit from the (potentially poisoned) global object, so we might see
```
erights: descriptors inherit from the (potentially poisoned) `Object.prototype`

shim/src/evaluators.js
```diff
@@ -4,70 +4,65 @@
...
+      // extra properties which weren't really there. Accessor properties have
+      // 'get/set/enumerable/configurable', while data properties have
+      // 'value/writable/enumerable/configurable'.
+      desc.configurable === true &&
```
erights: Both should be `=== false`

How did this work? We need tests that would have caught this.

shim/src/evaluators.js
```diff
@@ -4,70 +4,65 @@
...
+      desc.writable === true &&
+      //
+      // Check for accessor properties: we don't want to optimize these,
+      // they're obviously non-constant. Value properties can't have
+      // accessors at the same time, so this check is sufficient.
+      'value' in desc
```
erights: Unlike the previous code, this code does not deal with the case where `Object.prototype` has been poisoned to cause confusion when interpreting descriptors. Needs to be:

`&& !('get' in desc)` rather than `&& 'value' in desc`

Or we could (and probably should) write it less confusingly using explicit own property checks so we don't need to worry about inherited confusion.

shim/src/evaluators.js
```diff
@@ -4,70 +4,65 @@
...
+    );
   });
+
   return constants;
 }
 
 function buildOptimizer(constants) {
+  if (constants.length === 0) return '';
```
erights: How did the old code work without the `.length === 0` check?

jfparadis: It created an unused line:
```jf
const {} = arguments[0]; 
```

PR #151, opened 2018-07-07
Console.error tantrums

jfparadis:
Using console.warn() makes messages more visible.

### DIFF
--- a/shim/src/utilities.js
+++ b/shim/src/utilities.js
@@ -5,16 +5,18 @@
 
 export function throwTantrum(s, err = undefined) {
   const msg = `please report internal shim error: ${s}`;
+
   // note: we really do want to log these 'should never happen' things. there
   // might be a better way to convince the linter, though.
   // eslint-disable-next-line no-console
-  console.log(msg);
+  console.error(msg);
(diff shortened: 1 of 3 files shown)


shim/src/utilities.js
```diff
@@ -5,16 +5,18 @@
 
 export function throwTantrum(s, err = undefined) {
   const msg = `please report internal shim error: ${s}`;
+
   // note: we really do want to log these 'should never happen' things. there
   // might be a better way to convince the linter, though.
   // eslint-disable-next-line no-console
-  console.log(msg);
+  console.warn(msg);
```
ljharb: console.error, if they’re things that should never happen?

jfparadis: Do we need a distinction form the actual error thrown?

ljharb: i'm just referring to the console message - "warn" is more for stuff that's worth fixing but not necessary to fix, "error" is for stuff that's an error - and people's console filtering settings might be set to "only error". not a big deal either way, ofc.

jfparadis:
@ljharb Added tests yesterday.

PR #154, opened 2018-07-10
Conditional repair accessors

jfparadis:
Simpler implementation
Added more tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ node_js:
   - "10"
   - "9"
   - "8"
+  - "8.0"
   - "7"
   - "6"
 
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -3,6 +3,7 @@ node_js:
   - "10"
   - "9"
   - "8"
+  - "8.0"
```
jfparadis: Added node 8 which has the bug

shim/src/accessors.js
```diff
@@ -20,64 +20,57 @@
 // realms and it currently can't contain code coverage metrics.
 /* istanbul ignore file */
 export function repairAccessors() {
-  const {
-    getPrototypeOf,
-    defineProperty,
-    getOwnPropertyDescriptor,
-    prototype: objectPrototype
-  } = Object;
+  try {
```
jfparadis: Light detection.

PR #155, opened 2018-07-10
Guard against accessors on the unsafe global object

jfparadis:
Abort if an accessor is found on the unsafe global object instead of a
data property. We should never get into this non standard situation.


warner:
LGTM

PR #163, opened 2018-07-13
Fix optimizer

jfparadis:
Tests on the way

### DIFF
--- a/shim/src/evaluators.js
+++ b/shim/src/evaluators.js
@@ -11,6 +11,7 @@ import {
   getOwnPropertyDescriptors,
   getOwnPropertyNames,
   getPrototypeOf,
+  objectHasOwnProperty,
   regexpMatch,
   setPrototypeOf,
   stringIncludes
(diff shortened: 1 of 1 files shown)


shim/src/evaluators.js
```diff
@@ -19,42 +20,44 @@ import { ScopeHandler } from './scopeHandler';
...
 function getOptimizableGlobals(safeGlobal) {
   const descs = getOwnPropertyDescriptors(safeGlobal);
 
+  // getOwnPropertyNames does ignore Symbols so we don't need this extra check:
+  // typeof name === 'string' &&
   const constants = arrayFilter(getOwnPropertyNames(descs), name => {
-    // Ensure we have a valid identifier.
-    // getOwnPropertyNames does ignore Symbols so we don't need this extra check:
-    // typeof name === 'string' &&
+    // Ensure we have a valid identifier. We use 'apply' rather than
```
erights: We use `regexpMatch` rather than...

PR #165, opened 2018-07-14
Improve optimizer

jfparadis:
Move into own module
Fix bug with regexp
Ignore eval and keywords
@erights


shim/src/commons.js
```diff
@@ -48,5 +48,6 @@ export const objectHasOwnProperty = uncurryThis(Object.prototype.hasOwnProperty)
   arrayPop = uncurryThis(Array.prototype.pop),
   arrayJoin = uncurryThis(Array.prototype.join),
   arrayConcat = uncurryThis(Array.prototype.concat),
-  regexpMatch = uncurryThis(RegExp.prototype.match),
```
jfparadis: Bug: `RegExp.prototype.match` is `undefined`

PR #166, opened 2018-07-15
Improve evaluators

jfparadis:
Avoid getPrototypeOf in scope handler set
Use 'this' to bind optimizer constants
Safe Function(): tests found errors (tests disabled)
@erights

### DIFF
--- a/shim/src/scopeHandler.js
+++ b/shim/src/scopeHandler.js
@@ -33,10 +33,6 @@ export function createScopeHandler(unsafeRec) {
   // realm's code or if it is user-land invocation, so we can react differently.
   let useUnsafeEvaluator = false;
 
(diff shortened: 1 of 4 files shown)


shim/src/scopeHandler.js
```diff
@@ -26,17 +26,13 @@ const alwaysThrowHandler = new Proxy(freeze({}), {
  * - hide the unsafeGlobal which lives on the scope chain above the 'with'.
  * - ensure the Proxy invariants despite some global properties being frozen.
  */
-export function createScopeHandler(unsafeRec) {
+export function createScopeHandler(unsafeRec, safeGlobal = {}) {
```
erights: We should talk about whether this should have a safeGlobal arg, whether it should default to anything, and whether {} is the right default.

PR #167, opened 2018-07-20
Improve optimizer 3

jfparadis:
Block reserved null, true, false, this
Unblock await

### DIFF
--- a/shim/src/optimizer.js
+++ b/shim/src/optimizer.js
@@ -9,13 +9,23 @@ import {
 // todo: think about how this interacts with endowments, check for conflicts
 // between the names being optimized and the ones added by endowments
 
-// Admits many (but not all) legal variable names: starts with letter/_/$,
-// continues with letter/digit/_/$. It will reject many legal names that
-// involve unicode characters. \w is equivalent [a-zA-Z_0-9]
+/**
+ * Simplified validation of indentifier names: may only contain alphanumeric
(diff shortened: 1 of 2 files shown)


jfparadis:
@erights, reserved keywords `this`, `arguments` (strict mode), `true`, `false`, `null` are not clearly referenced in section 11.6.1 (Identifier Names). Also, `async` is allowed in browsers but should it be?

shim/src/optimizer.js
```diff
@@ -51,20 +61,27 @@ const keywords = new Set([
...
   'implements',
   'package',
   'protected',
   'interface',
   'private',
   'public',
 
-  // contextual worth refusing
-  'let',
-  'async',
```
erights: I think we should continue to omit `async` or anything similarly almost-reserved. It costs little to omit these, since no one will use them anyway. By omitting them, we have fewer worries we need to think about.

We should also omit `arguments` on the same grounds. See https://github.com/Agoric/TinySES/blob/master/src/tinyses.js#L49

jfparadis: Ok, re-omitting `async`, maybe those should be part of the specs?
FYI `arguments` is already omitted, and we have a test for it.

erights:
Oops. By "omit" I meant: Add them to this blacklist so they would be omitted from possible variable names. Sorry for the confusion!

PR #168, opened 2018-07-20
Test realm facade

jfparadis:
Add tests
Fix toString() for constructor and instance

### DIFF
--- a/shim/src/realmFacade.js
+++ b/shim/src/realmFacade.js
@@ -3,9 +3,6 @@ import { cleanupSource } from './utilities';
 // buildChildRealm is immediately turned into a string, and this function is
 // never referenced again, because it closes over the wrong intrinsics
 
-// todo: This function is stringified and evaluated outside of the primal
-// realms and it currently can't contain code coverage metrics.
-/* istanbul ignore file */
 export function buildChildRealm(BaseRealm) {
   const { initRootRealm, initCompartment, getRealmGlobal, realmEvaluate } = BaseRealm;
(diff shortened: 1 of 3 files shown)


jfparadis:
@erights, toString() changed to:
```
Realm.toString() => "function Realm() { [shim code] }" 
Realm.makeCompartment().toString() => "[object Realm]"
```

shim/src/realmFacade.js
```diff
@@ -99,13 +96,20 @@ export function buildChildRealm(BaseRealm) {
     }
   }
 
-  defineProperty(Realm.prototype, 'toString', {
+  defineProperty(Realm, 'toString', {
```
ljharb: do you control the Function toString value such that `Function.toString.call` will respect this property? What happens if i `delete Realm.toString; String(Realm)`?

erights: It will print the source code from the shim, which thereby fails to perfectly emulate the specified behavior. For now, I don't think we should fix this, but rather document it as a low priority open issue.

This all gets more interesting with the functions we write for purposes of stringifying and then evaluating in another realm. Separately from this PR, we should examine attacks by forced divergences among ways to stringify those functions.

jfparadis: Correct, this is not done for security but strictly for ergonomics, to help with debugging.

ljharb: Overriding Function.prototype.toString would provide the security though, no?

PR #169, opened 2018-07-20
Improve Realm creation

jfparadis:
Use assert() to guard the weakmaps used as private fields
Move TypeError on new Realm() to the constructor

### DIFF
--- a/shim/src/realm.js
+++ b/shim/src/realm.js
@@ -6,6 +6,7 @@ import {
   createSafeEvaluatorWhichTakesEndowments,
   createFunctionEvaluator
 } from './evaluators';
+import { assert } from './utilities';
 import { create, defineProperty, defineProperties, freeze, arrayConcat } from './commons';
 
 // Create a registry to mimic a private static members on the realm classes.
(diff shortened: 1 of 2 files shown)


shim/src/realm.js
```diff
@@ -14,30 +15,20 @@ import { create, defineProperty, defineProperties, freeze, arrayConcat } from '.
 const UnsafeRecForRealmClass = new WeakMap();
 
 function getUnsafeRecForRealmClass(RealmClass) {
-  if (Object(RealmClass) !== RealmClass) {
-    // Detect non-objects.
-    throw new TypeError('internal error: bad object, not a Realm constructor');
```
jfparadis: Preventing new Realm() via this side effect felt incorrect. What if there was no weakmap().

shim/src/realmFacade.js
```diff
@@ -69,14 +70,29 @@ export function buildChildRealm(BaseRealm) {
...
   class Realm {
+    constructor() {
+      // The Realm constructor is not intended to be used with the new operator
+      // or to be subclassed. It may be used as the value of an extends clause
+      // of a class definition but a super call to the Realm constructor will
+      // cause an exception.
+
+      // When Realm is called as a function, an exception is also raised because
+      // a class constructor cannot be invoked without 'new'.
+      throw new TypeError('Realm is not a constructor');
```
jfparadis: @erights I've modeled this on the specs of Symbol.

PR #178, opened 2018-07-28
Small items: clarification of code and comments, and removal of unecessary code.

jfparadis:


jfparadis:
@erights, I am addressing your suggestions here, and proposing a few changes. You can review the five commits individually.

PR #179, opened 2018-07-29
Added tests for unsafeRec

jfparadis:


shim/test/module/unsafeRec.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+  };
+
+  const unsafeGlobal = createNewUnsafeGlobalForBrowser();
+
+  t.equal(unsafeGlobal, window);
+  t.equal(iframe.style.display, 'none');
+  t.deepEqual(body, [iframe]);
+
+  delete global.window;
```
erights: Put these deletes in a `finally` clause

shim/test/module/unsafeRec.js
```diff
@@ -0,0 +1,50 @@
...
+  t.deepEqual(Object.keys(sharedGlobalDescs), Object.keys(expectedSharedGlobalDescs));
+  t.equal(unsafeEval, unsafeGlobal.eval);
+  t.equal(unsafeFunction, unsafeGlobal.Function);
+  t.deepEqual(allShims, []);
+});
+
+test('createNewUnsafeGlobalForBrowser', t => {
+  t.plan(3);
+
+  const window = vm.runInNewContext('this');
```
erights: Let's see if I have this right:

The code in this test case assumes (without checking) that it is run on node. It then provides a minimal emulation of a browser environment in order to test browser-specific code when run on node.

What happens when this test is run on a browser? Should we package this emulation up in a testing utils library of some sort? Should we also have a minimal node emulation that runs in the browser, so that we can test node specific code when running tests on the browser?

In any case, it is worth a comment to explain your intentions and assumptions here.

shim/test/realm/test-repair.js
```diff
@@ -25,7 +25,7 @@ function testForBug(o) {
 }
 
 test('fix the bug in which accessor methods leak the global', t => {
-  const unfixedGlobal = getNewUnsafeGlobal();
+  const unfixedGlobal = vm.runInNewContext('this');
```
erights: The old code was host independent. This seems better than the new code.

jfparadis: True. However, the old code used the shim to create an artifact (unfixedGlobal) for testing without testing it, and increased code coverage artifically. 

I have not added browser testing yet, however I will fix this in preparation.

shim/test/module/unsafeRec.js
jfparadis: Currentlty, these tests only execute on node. Typically, npm libraries have unit tests running on node or on node + jsdom. We can't use jsdom because it doesn't support iframe.contentWindow().

In my mind, it would be beneficial to have a complete test coverage whether we develop on node or on the browser, withouth requiring to execute all tests on both platforms to detect a breakage.

I have not added browser testing yet, however I have added a guard in preparation.

shim/src/unsafeRec.js
```diff
@@ -31,7 +31,7 @@ function createNewUnsafeGlobalForNode() {
   return unsafeGlobal;
 }
 
-function createNewUnsafeGlobalForBrowser() {
+export function createNewUnsafeGlobalForBrowser() {
```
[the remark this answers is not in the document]
ljharb: you'd need a decorator for that ;-)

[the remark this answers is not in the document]
ljharb: @jfparadis this wouldn't be exclusive to tape; you'd use a decorator that made a private thing public (or made an exported `undefined` binding be the actual function, or something) when `process.env.NODE_ENV` was `'test'`.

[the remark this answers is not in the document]
jfparadis: @ljharb, I agree, it's not exclusive to `tape`, and I'm wondering one of the many extensions could provide the behaviour we need.

The main issue: after we invoke our methods that repair the environment, using decorators or not, there is no complete revert, and tests can easily bleed into all subsequent calls. We could of course develop a "capture & restore" plugin for the global and its intrinsics, but it's much simpler to start with a new realm. This is an area where test262 is stronger, every test being executed in a new realm with a fresh set of intrinsics.

In my mind, it's more important to solve this issue with the execution environment, and doing so would automatically resolve all other issues: we simply load the class in two contexts, and then check the correct behaviour without export or decorators.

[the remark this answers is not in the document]
jfparadis: it looks like `multi-tape` was designed to do what we need:

> Since it runs all tests in a single node instance, global variables and mocks will spread between the tests in an uncontrolled manner. `multi-tape` solves this by running each js-file in a separate node-process.
https://www.npmjs.com/package/multi-tape

PR #180, opened 2018-07-31
Use "defineProperties" instead of "defineProperty" 

jfparadis:
... to get throwing behavior for free.
@erights Should work as discussed.


shim/src/commons.js
```diff
@@ -9,9 +9,8 @@
 export const {
   assign,
   create,
-  defineProperties,
-  defineProperty,
   freeze,
+  defineProperties, // Object.defineProperties is allowed to fail silentlty.
```
warner: typo in comment

shim/src/repair/accessors.js
```diff
@@ -59,48 +60,46 @@ export function repairAccessors() {
     return obj;
   }
 
-  defineProperties(objectPrototype, {
```
erights: I think the original, with `defineProperties`, is better than the revision, with `definePropertyOrThrow`. Remember, we're only overcoming an irregularity with `Object.defineProperty` specifically, that `defineProperties` does not suffer from.

jfparadis: I dropped this for a while since I wasn't able to find the irregularity in the specs:
```
19.1.2.4 Object.defineProperty

    If Type(O) is not Object, throw a TypeError exception.
    Let key be ? ToPropertyKey(P).
    Let desc be ? ToPropertyDescriptor(Attributes).
    Perform ? DefinePropertyOrThrow(O, key, desc).
    Return O. 
```
The spec matches the behaviour I see in node and all evergreen browsers. Was the issue restricted to WindowProxy?
https://esdiscuss.org/topic/figuring-out-the-behavior-of-windowproxy-in-the-face-of-non-configurable-properties

PR #195, opened 2019-01-17
fix #188

errorx666:


errorx666:
Tests 341 and 342 are also failing with my `master` branch, which has no changes.

caridy:
The polyfill was moved to https://github.com/Agoric/realms-shim, and this repo is not exclusively tracking the spec text.

PR #224, opened 2020-02-26
closes #183: `evaluation(sourceText, options)` to support evaluating modules and scripts

caridy:


spec/index.emu
```diff
@@ -38,19 +38,35 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
   <h1>Realm Objects</h1>
   <emu-clause id="sec-realm-abstract-operations">
     <h1>Realm Abstract Operations</h1>
-    <emu-clause id="sec-performRealmEvaluation" aoid="PerformRealmEvaluation">
-      <h1>PerformRealmEvaluation ( _x_, _evalRealm_ )</h1>
+
+    <emu-clause id="sec-perform-realm-script-evaluation" aoid="PerformRealmScriptEvaluation">
```
caridy: just renaming things to match the new abstract operation for modules

spec/index.emu
```diff
@@ -90,16 +106,25 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
...
+          1. Let _sourceType_ be *undefined*.
+          1. If _options_ is not *undefined*, then
+            1. Let _options_ be ? ToObject(_options_).
+            1. Let _sourceType_ be ? Get(_options_, *"type"*).
+          1. If _sourceType_ is *"module"*, then
+            1. Return ? PerformRealmModuleEvaluation(_sourceText_, _realm_).
+          1. Else if _sourceType_ is *"script"*, then
+            1. Return ? PerformRealmScriptEvaluation(_sourceText_, _realm_).
+          1. Else
+            1. Throw a RangeError exception.
```
caridy: @erights @ljharb for now I'm assuming that we want to throw if `options.type` is not provided or is invalid value. We can definitely relax this, and add a default value. Any thoughts?

```js
const r = new Realm();
r.evaluate(`1`, { type: 'module' });
r.evaluate(`2`, { type: 'script' });
```

spec/index.emu
```diff
@@ -38,19 +38,35 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
   <h1>Realm Objects</h1>
   <emu-clause id="sec-realm-abstract-operations">
     <h1>Realm Abstract Operations</h1>
-    <emu-clause id="sec-performRealmEvaluation" aoid="PerformRealmEvaluation">
-      <h1>PerformRealmEvaluation ( _x_, _evalRealm_ )</h1>
+
+    <emu-clause id="sec-perform-realm-script-evaluation" aoid="PerformRealmScriptEvaluation">
+      <h1>PerformRealmScriptEvaluation ( _sourceText_, _realm_ )</h1>
```
caridy: this operation matches https://tc39.es/ecma262/#sec-scriptevaluationjob

spec/index.emu
```diff
@@ -38,19 +38,35 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
...
+      1. Let _s_ be ParseScript(_sourceText_, _realm_, _hostDefined_).
       1. If _s_ is a List of errors, then
         1. Perform HostReportErrors(_s_).
         1. Return NormalCompletion(*undefined*).
       1. Return ? ScriptEvaluation(_s_).
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-perform-realm-module-evaluation" aoid="PerformRealmModuleEvaluation">
+      <h1>PerformRealmModuleEvaluation ( _sourceText_, _realm_ )</h1>
```
caridy: this operation must match https://tc39.es/ecma262/#sec-toplevelmoduleevaluationjob

spec/index.emu
ljharb: seems better to keep it explicit, but at that point why have it in the options bag instead of as a required second argument?

caridy: good question! two main reasons that we have discussed: 

* `evaluate()` is also a thing in evaluator/compartment, just with more stuff on it, like endowments.
* maybe in the future we will have more options here, like strict or sloppy, etc.

But I'm fine either way. We can overload the arguments if we need more options in the future. @erights any opinion here?

ljharb: I'd generally expect options bags to only hold optional arguments; in the case of "evaluate", the parse goal it's evaluating in seems pretty first-class to me. In other words, I'd expect an optional options bag (now or in the future) to be the *third* argument.

littledan:
What's the motivation for this PR?

littledan:
Note, I think we'll need some host integration here. Right now, hosts can assume that all module evaluation starts with them (since currently, hosts have to provide modules to ES), so they have some time to set modules up and add appropriate metadata, e.g., so that dynamic `import()` inside them works. If JS itself is creating modules, then we should consider adding a host hook to allow the appropriate logic to be inserted.

spec/index.emu
caridy: feedback from the SES meeting yesterday about this question:

* majority of the group will like to have a default value. I still think being explicit about this is a safer (it is more restrictive, and can be relaxed in the future).
* default value to `strict-script`, which is kinda a new concept, their argument was to not favor sloppy mode, which is still possible, but only via `eval` from `globalThis`. No capability lost there, just less convience.
* if there is a default value, then it is better to have the options bag. which I believe all agree, but first lets get to some consensus about that first question.
* there was some discussion about tc53 use cases, and whether or not some folks will simply be able to throw when `type=module` is provided, we need more info about this use-case, and how that impact the API.

spec/index.emu
```diff
@@ -38,19 +38,35 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
...
+      <emu-alg>
+      1. Assert: _sourceText_ is an ECMAScript source text (see clause <emu-xref href="#sec-ecmascript-language-source-code"></emu-xref>).
+      1. Assert: _realm_ is a Realm Record.
+      1. Let _hostDefined_ be _evalRealm_.[[HostDefined]].
+      1. Let _m_ be ParseModule(_sourceText_, _realm_, _hostDefined_).
+      1. If _m_ is a List of errors, then
+        1. Perform HostReportErrors(_m_).
+        1. Return NormalCompletion(*undefined*).
+      1. Assert: All dependencies of _m_ have been transitively resolved and _m_ is ready for evaluation.
+      1. Return ? _m_.Evaluate().
```
littledan: You probably don't want to replace _m_.Evaluate(), as this is undefined for all Cyclic Module records (which this will be). ([spec](https://tc39.es/ecma262/#sec-moduleevaluation)).

Instead, maybe you want to perform ? _m_.Evaluate() and then return GetModuleNamespace(_m_).

spec/index.emu
```diff
@@ -38,19 +38,35 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
...
       1. Return ? ScriptEvaluation(_s_).
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-perform-realm-module-evaluation" aoid="PerformRealmModuleEvaluation">
+      <h1>PerformRealmModuleEvaluation ( _sourceText_, _realm_ )</h1>
+      <emu-alg>
+      1. Assert: _sourceText_ is an ECMAScript source text (see clause <emu-xref href="#sec-ecmascript-language-source-code"></emu-xref>).
+      1. Assert: _realm_ is a Realm Record.
+      1. Let _hostDefined_ be _evalRealm_.[[HostDefined]].
```
littledan: s/_evalRealm_/_realm_/ ?

spec/index.emu
```diff
@@ -38,19 +38,35 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
...
+      1. Let _m_ be ParseModule(_sourceText_, _realm_, _hostDefined_).
```
littledan: After parsing and before evaluating, did you want to link this module?

spec/index.emu
erights: > @erights did I miss anything?

That's what I remember, yes. For reference, the mtg is at https://www.youtube.com/watch?v=hqKGKDw-FCY&list=PLzDw4TTug5O1jzKodRDp3qec8zl88oxGd

spec/index.emu
zenparsing: As defined [here](https://tc39.es/ecma262/#sec-moduleevaluation), `Evaluate` will return `undefined` for non-throwing cases, correct?

Aren't users going to prefer to get a promise for a namespace object, or something like that, instead? If so, are we sure that we want to put both module and script evaluation into the same method?

spec/index.emu
littledan: I don't think this should be taken as a requirement. That abstract operation is going away in https://github.com/tc39/ecma262/pull/1597/files#diff-3540caefa502006d8a33cb1385720803L24193 , which TC39 has reached consensus on.

spec/index.emu
littledan: Agreed; see wording suggestion in https://github.com/tc39/proposal-realms/pull/224#discussion_r385806892

spec/index.emu
caridy: good catch! missed that line.

spec/index.emu
caridy: once that's merged, we can adjust here I think.

spec/index.emu
zenparsing: With top-level-await, this will return a promise, correct? Does it seem strange to have a method return a promise or not depending on the presence of a property in an option bag (i.e. `type: 'module'` or `type: 'script'`)?

caridy: fixed!

spec/index.emu
littledan: Did the SES meeting consider
- Using separate method names, as these really have different types, and for the Promise vs not issue @zenparsing identified?
- Having an expression form, as @erights initially proposed at the start of the thread?

spec/index.emu
littledan: @zenparsing This is a really good point. It's starting to feel to me like these should be different method names, as @erights originally proposed in #183.

Given that TLA is already at Stage 3, should we make this a Promise pre-emptively? It'd be good to not have open issues like this settled on before Stage 3.

ljharb: Scripts run synchronously; i shouldn't be forced to wait a tick to know that my script evaluated.

littledan: @ljharb I don't think anyone is proposing that script evaluation return a Promise.

guybedford: @ljharb with TLA, even though the script runs synchronously (you don't have to wait a tick for the execution to start), it will still return a promise which represents any async completion in the case of there being an await in the graph. If there was no await, the TLA spec could possibly be adjusted to eg return _undefined_ completion instead of an empty resolved promise, but I believe the v8 API will just always return a promise currently.

spec/index.emu
littledan: This wording seems fine to me, FWIW.

spec/index.emu
littledan: It's not really about what the V8 API is (that could change)--it's more a general design principle, that we shouldn't have APIs that sometimes return Promises and sometimes not.

littledan: I think we should consider this issue that @zenparsing raised as unresolved, and resolve it before Realms goes to Stage 3. (Some other options would be to punt on module evaluation, defering things to `realm.globalThis.eval`, as I proposed in #226, or to simply close this PR and leave `evaluate` to script evaluation.)

caridy:
closing in favor of https://github.com/tc39/proposal-realms/pull/235

PR #233, opened 2020-03-04
Normative: Run HostImportModuleDynamically in the correct Realm

littledan:


spec/index.emu
```diff
@@ -119,13 +119,24 @@ location: https://rawgit.com/tc39/proposal-realms/master/index.html
...
+          1. Remove _newContext_ from the execution context stack and restore _callerContext_ as the running execution context.
           1. Return _promiseCapability_.[[Promise]].
           </emu-alg>
 
           <emu-note>
               Extensible web: This is equivalent to dynamic import without having to evaluate a script source, which might not be available (e.g.: when CSP is blocking source evaluation).
           </emu-note>
+
+          <emu-note type=editor>
+              An extra execution context is created around the call to HostImportModuleDynamically in order to execute within the appropriate Realm--the host can reference the current Realm to figure out how the import should work. An alternative phrasing would be to pass the Realm as a paramter to HostImportModuleDynamically, which could be revisited when merging into the main specification.
```
ljharb:
```suggestion
              An extra execution context is created around the call to HostImportModuleDynamically in order to execute within the appropriate Realm--the host can reference the current Realm to figure out how the import should work. An alternative phrasing would be to pass the Realm as a parameter to HostImportModuleDynamically, which could be revisited when merging into the main specification.
```

PR #242, opened 2020-04-27
fixes #241: adding Realm.prototype.import docs

### LINKED ISSUE #241: missing `import()` method in API(TypeScript)?
In the latest spec,
https://rawgit.com/tc39/proposal-realms/master/index.html#sec-realm.prototype.import

`Realm.prototype.import(specified)` is defined, but not in API(TypeScript) in README:
https://github.com/tc39/proposal-realms#api-typescript-format

caridy:


caridy:
cc @zaki-yama

README.md
```diff
@@ -97,12 +97,22 @@ class FakeWindow extends Realm {
...
 ```
 
+### Example: Importing Module
+
+```js
+let r = new Realm();
+const { x } = await r.import('/path/to/foo.js');
+```
+
+In this example, the new realm will fetch, and evaluate the module, and extract the `x` named export from that module namespace. `Realm.prototype.import` is equivalent to the dynamic import syntax (e.g.: `const { x } = await r.import('/path/to/foo.js');` from within the realm. In some cases, evaluation will not be available (e.g.: in browsers, CSP might block unsafe-eval), while importing from module is still possible.
```
erights, line 107: Your example code is not the example you're looking for. `r.import` should be `import`

PR #250, opened 2020-05-27
[Editorial] Enable rendered pages through GH Actions

leobalter:


caridy:
I think there are some steps at the end of README that probably need some update to open the local version. Can you double check? Also the history of the .emu file was lost as part of this PR, maybe moving the file could have preserved the years of history of the evolution of the spec. Not a big deal though.

leobalter:
I'm following up the work with the readme and explainer files and I should have updates today.

My bad on the file history. I'm pretty sure I did `git mv` the file but ended up unlinking it by mistake. I could rework this through a forced push to master but I'm not sure if it would be good as well.

PR #251, opened 2020-05-27
Draft for the Explainer

leobalter:
This is an initial explainer that might and should be expanded. 

I believe it can be a nice starting point to land on the main branch so we follow up with specialized patches.


leobalter:
I'll merge this for now, we can follow up with specific fixes.

PR #253, opened 2020-05-28
fix the location of spec text in README

zaki-yama:
It seems the location of the spec text and the built html was changed in #250 .


leobalter:
please observe #251. It should be merged today.

zaki-yama:
@leobalter 
Oh, I'm sorry. Thank you for telling.

PR #263, opened 2020-06-23
Editorial: Use dotted intrinsics

ExE-Boss:
Legacy intrinsic aliases will be removed in <https://github.com/tc39/ecma262/pull/2056>.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12,32 +12,30 @@ location: https://tc39.es/proposal-realms/
 <script src="ecmarkup.js"></script>
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.4/styles/github.min.css">
 
-<emu-table id="table-1" caption="Well-known Intrinsic Objects">
-	<table>
-		<tbody>
-			<tr>
-				<th>Intrinsic Name</th>
-				<th>Global Name</th>
-				<th>ECMAScript Language Association</th>
-			</tr>
-			<tr>
-				<td>%Realm%</td>
-				<td><code>Realm</code></td>
-				<td>…</td>
-			</tr>
-			<tr>
-				<td>%RealmPrototype%</td>
-				<td><code>Realm.prototype</code></td>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -12,32 +12,30 @@
...
-			<tr>
-				<td>%RealmPrototype%</td>
-				<td><code>Realm.prototype</code></td>
-				<td>…</td>
-			</tr>
-		</tbody>
-	</table>
-</emu-table>
+<emu-clause id="sec-well-known-intrinsic-objects">
+  	<h1>Well-known intrinsic objects</h1>
```
leobalter, line 16: nit: indentation is inconsistent

spec.html
```diff
@@ -12,32 +12,30 @@
...
+	<emu-table id="table-7" caption="Well-known Intrinsic Objects">
```
leobalter, line 17: This is Table 8 in ECMA-262 today. It might be easier leaving this as Table 1 or a non number identifier.

ljharb:
It doesn't need to be merged for this to land; the dotted intrinsic form landed a long time ago, and is preferred.

spec.html
ExE-Boss: This is so that **ecmarkup** auto‑links `<emu-xref href="#table-7" />` in this specification as local links, instead of links to <https://tc39.es/ecma262/#table-7>.

leobalter: I see. The id remains with the wrong number.

@tc39/ecma262-editors Is this something we can change in ECMA-262? I mean, having a nominal id rather than numeric?

leobalter:
It should be good enough to land, but I'd like @caridy's eyes here, which should happen this week.

spec.html
ljharb: We could change the legacy ID (and preserve it via an oldid) but it doesn't seem particularly important to change to me.

leobalter: Well, let's keep the off-by-one then. :)

PR #267, opened 2020-07-26
New explainer before a TAG review

leobalter:
changes are better seen in the different commits.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -61,6 +61,7 @@ Sites like Salesforce.com make extensive use of same-origin iframes to create su
 - Frameworks would be able to better craft the available API within the global object of the Realm, aiming for what is necessary to evaluate the code that might need no access or just limited access to the DOM or other Web APIs.
 - Tailoring up the exposed set of APIs into the code within the Realm provides a better developer experience for a less expensive work compared to tailoring down a full set of exposed APIs - e.g. iframes - that includes handling presence of `[LegacyUnforgeable]` attributes like `Window.top`.
 - We hope the usage of Realms will be somewhat lighter weight (both in terms of memory and CPU) for the browser if compared to iframes, especially when frameworks rely on several Realms in the same application.
+- Realms are not accessible from by traversing the DOM of the incubator realm. This will ideal and/or better approach compared to attaching iframes elements and their contentWindow to the DOM. Detaching iframes would even add a new own set of problems.
 
 Realms are complementary to stronger isolation mechanisms such as Workers and cross-origin iframes. They are useful for contexts where synchronous execution is an essential requirement, e.g., emulating the DOM for integration with third-party code. Realms avoid often-prohibitive serialization overhead by using a common heap to the surrounding context.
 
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -50,15 +50,21 @@
...
 
-Code can be evaluated and executed within the Realm's Environment Record and Execution Context.
+Same-origin iframes also create a new global object which is synchronously accessible. Realms differ from same-origin iframes by omitting Web APIs such as the DOM.
 
-The Realms API offers a way to import modules asynchronously, just like the `import()` expression, following the same design patterns. It does not restrict code to be synchronously executed through regular evaluation built-ins (e.g. `eval` and `Function`). The Realms will not behave like different [Agents](https://tc39.es/ecma262/#sec-agents). They do not create - neither rely on - multi-threading.
+Sites like salesforce.com make extensive use of same-origin iframes to create such global objects. Our experience with same-origin iframes motivated us to create this proposal, which has the following advantages:
 
-Any code executed within a Realm may introduce changes to the Realm global variables or built-ins, limited to the Realm's Execution Context.
+We hope that it will be somewhat lighter weight (both in terms of memory and CPU) for the browser to create new Realms than iframes.
```
littledan, line 61: Should these be bullet points?

leobalter: yes

explainer.md
```diff
@@ -181,6 +187,18 @@ const { doSomething } = await realm.import('./file.js');
...
 The Realms API does not introduce a new way to evaluate code, it is subject to the existing evaluation mechanisms such as the [Content-Security-Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy).
 
+## <a name='Whynotseparateprocesses'></a>Why not separate processes?
+
+Creating a Realm that runs in a separate process is another alternative, while allowing users to define and create their own protocol of communication between these processes.
+
+This alternative was discarded for two main reasons:
+
+1. There are existing mechanism to achieve this today in both browsers, and nodejs. E.g.: cross domain iframes, workers, etc. They seem to be good enough when asynchronous communication is sufficient to implement the feature.
+2. Asynchronous communication is a deal-breaker for many use-cases, and sometimes it just added complexity for cases where a same-process Realm is sufficient.
```
caridy, line 197: we should add to this bullet that this is possible when security is NOT an issue, to avoid getting into this controversy around security. A good example here is AMP, they run in a cross domain iframe, they just want to have more control about what code they executed in that cross domain app.

explainer.md
```diff
@@ -50,15 +50,23 @@
...
+Sites like Salesforce.com make extensive use of same-origin iframes to create such global objects. Our experience with same-origin iframes motivated us to steer this proposal forward, which has the following advantages:
 
-Any code executed within a Realm may introduce changes to the Realm global variables or built-ins, limited to the Realm's Execution Context.
+- Frameworks would be able to better craft the available API within the global object of the Realm, aiming for what is necessary to evaluate the code that might need no access or just limited access to the DOM or other Web APIs.
+- Tailoring up the exposed set of APIs exposed to the code within the Realm provides a better developer experience for a less expensive work compared to tailoring down a full set of exposed APIs - e.g. iframes - that includes handling presence of `[LegacyUnforgeable]` attributes like `Window.top`.
+- We hope the usage of Realms will be somewhat lighter weight (both in terms of memory and CPU) for the browser if compared to iframes, especially when frameworks rely on several Realms in the same application.
```
caridy, line 63: I think we can add here that iframes are always accessible from the DOM in the incubator call by traversing the DOM, accessing the iframe element, and accessing its contentWindow, which is not ideal. Also, adding that the iframes are detacheable, which has its own set of problems.

PR #269, opened 2020-08-06
Expand the example for the FakeDocument

leobalter:


leobalter:
This PR had an interactive feedback from @caridy and @littledan. Thanks for all the input!

PR #279, opened 2020-11-05
Editorial: Create ValidateRealmObject abstraction

leobalter:
Fixes #278

cc @ljharb


leobalter:
The reason I did not call this IsRealm is the fact I'm not returning anything from the abstraction. This sounds more straightforward but I'd be happy to change it.

ljharb (approved):
It might also be good to update https://tc39.es/proposal-realms/#sec-properties-of-the-realm-prototype-object so indicate that it does not have either slot.

ljharb (approved):

PR #283, opened 2020-11-08
Editioral: rename the host hook to HostInitializeSyntheticRealm

littledan:
The term "synthetic realm" is proposed to be used in the HTML/Realm
integration PR to describe the contrast with the main "principal"
Realm. This PR renames the host hook to match, and to avoid the
confusion noted in #281


Jack-Works:
there is merging conflict 👀

caridy (approved):

PR #291, opened 2021-03-25
[draft] first draft for isolated realms

caridy:
Spec'ing the result of the discussion from https://github.com/tc39/proposal-realms/issues/289 to describe the isolated realms.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32,8 +32,118 @@ location: https://tc39.es/proposal-realms/
 	</emu-table>
 </emu-clause>
 
+<emu-clause id="sec-wrapped-function-exotic-objects">
+	<h1>Wrapped Function Exotic Objects</h1>
+	<p>A wrapped function exotic object is an exotic object that wraps a callable object. A wrapped function exotic object is callable (it has a [[Call]] internal method). Calling a wrapped function exotic object generally results in a call of its wrapped function.</p>
+
+	<p>An object is a <dfn id="wrapped-function-exotic-object">wrapped function exotic object</dfn> if its [[Call]] internal method use the following implementations, and its other essential internal methods use the definitions found in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. These methods are installed in WrappedFunctionCreate.</p>
+
+	<p>Wrapped function exotic objects do not have the internal slots of ECMAScript function objects listed in <emu-xref href="#table-internal-slots-of-ecmascript-function-objects"></emu-xref>. Instead they have the internal slots listed in <emu-xref href="#table-internal-slots-of-wrapped-function-exotic-objects"></emu-xref>, in addition to [[Prototype]] and [[Extensible]].</p>
+	<emu-table id="table-internal-slots-of-wrapped-function-exotic-objects" caption="Internal Slots of Wrapped Function Exotic Objects" oldids="table-28">
+	<table>
+		<tbody>
+		<tr>
+		<th>
+			Internal Slot
+		</th>
+		<th>
+			Type
+		</th>
+		<th>
+			Description
+		</th>
+		</tr>
+		<tr>
+		<td>
+			[[WrappedTargetFunction]]
+		</td>
+		<td>
+			Callable Object
+		</td>
+		<td>
+			The wrapped function object.
+		</td>
+		</tr>
+		<tr>
+		<td>
+			[[Realm]]
+		</td>
+		<td>
+			Realm Record
+		</td>
+		<td>
+			The realm in which the wrapped function object was created.
+		</td>
+		</tr>
+		</tbody>
+	</table>
+	</emu-table>
+
+	<emu-clause id="sec-wrapped-function-exotic-objects-call-thisargument-argumentslist">
+	<h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
+	<p>The [[Call]] internal method of a wrapped function exotic object _F_ takes arguments _thisArgument_ (an ECMAScript language value) and _argumentsList_ (a List of ECMAScript language values). It performs the following steps when called:</p>
+	<emu-alg>
+		1. Let _target_ be _F_.[[WrappedTargetFunction]].
+		1. Assert: IsCallable(_target_) is *true*.
+		1. Let _targetRealm_ be ? GetFunctionRealm(_target_).
+		1. Let _callerRealm_ be ? GetFunctionRealm(_F_).
+		1. For each element _key_ of _argumentsList_, do
+			1. Let o be _argumentsList_[_key_].
+			1. Set _argumentsList_[_key_] to ? GetWrappedValue(_targetRealm_, _o_).
+		1. Let _wrappedThisArgument_ to ? GetWrappedValue(_targetRealm_, _thisArgument_).
+		1. Let _value_ be ? Call(_target_, _wrappedThisArgument_, _argumentsList_).
+		1. Return ? GetWrappedValue(_callerRealm_, _value_).
+	</emu-alg>
+	</emu-clause>
+
+	<emu-clause id="sec-wrappedfunctioncreate" aoid="WrappedFunctionCreate">
+	<h1>WrappedFunctionCreate ( _callerRealm_, _targetFunction_ )</h1>
+	<p>The abstract operation WrappedFunctionCreate takes arguments _callerRealm_ and _targetFunction_. It is used to specify the creation of new wrapped function exotic objects. It performs the following steps when called:</p>
+	<emu-alg>
+		1. Assert: _callerRealm_ is a Realm Record.
+		1. Assert: IsCallable(_targetFunction_) is *true*.
+		1. Let _internalSlotsList_ be the internal slots listed in <emu-xref href="#table-internal-slots-of-wrapped-function-exotic-objects"></emu-xref>, plus [[Prototype]] and [[Extensible]].
+		1. Let _obj_ be ! MakeBasicObject(_internalSlotsList_).
+		1. Set _obj_.[[Prototype]] to %Function.prototype% from _callerRealm_.
+		1. Set _obj_.[[Call]] as described in <emu-xref href="#sec-wrapped-function-exotic-objects-call-thisargument-argumentslist"></emu-xref>.
+		1. Set _obj_.[[Realm]] to _callerRealm_.
+		1. Set _obj_.[[WrappedTargetFunction]] to _targetFunction_.
+		1. Return _obj_.
+	</emu-alg>
+	</emu-clause>
+</emu-clause>
+
 <emu-clause id="sec-realm-objects">
 	<h1>Realm Objects</h1>
+	<emu-clause id="sec-realm-abstracts">
+		<h1>Realm Abstract Operations</h1>
+
+		<emu-clause id="sec-performrealmevaluation" aoid="PerformRealmEvaluation">
+			<h1>PerformRealmEvaluation ( _x_, _evalRealm_ )</h1>
+			<emu-alg>
+				1. Assert: Type(_x_) is String.
+				1. Assert: _evalRealm_ is a Realm Record.
+				1. Let _hostDefined_ be _evalRealm_.[[HostDefined]].
+				1. Let _s_ be ParseScript(_x_, _evalRealm_, _hostDefined_).
+				1. If _s_ is a List of errors, then
+					1. Perform HostReportErrors(_s_).
+					1. Return NormalCompletion(*undefined*).
+				1. Return ? ScriptEvaluation(_s_).
+			</emu-alg>
+		</emu-clause>
+
+		<emu-clause id="sec-getwrappedvalue" aoid="GetWrappedValue">
+			<h1>GetWrappedValue ( _callerRealm_, _value_ )</h1>
+			<emu-alg>
+				1. Assert: _callerRealm_ is a Realm Record.
+				1. If Type(_value_) is Object, then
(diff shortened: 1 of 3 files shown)


spec.html
```diff
@@ -187,6 +306,33 @@ <h1>Runtime Semantics: HostInitializeSyntheticRealm ( _realm_ )</h1>
...
+			<p>The host-defined abstract operation HostImportModuleBindingDynamically takes arguments _referencingScriptOrModule_ (a Script Record or Module Record or *null*), _specifier_ (a |ModuleSpecifier| String), _bindingName_ (a |IdentifierName| String) and _promiseCapability_ (a PromiseCapability Record). It performs any necessary setup work in order to make available the module corresponding to _specifier_ occurring within the context of the script or module represented by _referencingScriptOrModule_. _referencingScriptOrModule_ is *null*. It then performs FinishDynamicImport to finish the dynamic import process.</p>
+			<p>The implementation of HostImportModuleBindingDynamically must conform with the implementation of HostImportModuleDynamically first and foremost, second, it should conform with the following:</p>
+			<li>
+				The host environment must conform to one of the two following sets of requirements:
+				<dl>
+				<dt>Success path</dt>
+
+				<dd>
+					<ul>
+						<li>The completion value must be the exported value for the _bindingName_ associated to the Module Namespace Object.</li>
```
leobalter, line 322: This needs to add a step to include a resolution of `GetWrappedValue` over the completion value. It might have a promise wrapping the resolution of `FinishDynamicImport` to capture the values.

It might need to dupe the received promiseCapability to unwrap the binding values from the resolved namespace.

It also needs to guarantee the error is from this incubator realm.

spec.html
```diff
@@ -187,6 +306,33 @@ <h1>Runtime Semantics: HostInitializeSyntheticRealm ( _realm_ )</h1>
...
+					</ul>
+				</dd>
+
+				<dt>Failure path</dt>
+
+				<dd>
+					<ul>
+						<li>At some future time, the Module Namespace Object does not contain the _bindingName_ as part of the ExportNames, with the abrupt completion representing the cause of failure.</li>
```
leobalter, line 330: Same guarantees here for the error being part of this incubator realm.

leobalter (changes requested):

spec.html
caridy: you're correct, that's missing.

spec.html
```diff
@@ -32,8 +32,118 @@ <h1>Well-known intrinsic objects</h1>
...
+	</emu-clause>
+</emu-clause>
+
 <emu-clause id="sec-realm-objects">
 	<h1>Realm Objects</h1>
+	<emu-clause id="sec-realm-abstracts">
+		<h1>Realm Abstract Operations</h1>
+
+		<emu-clause id="sec-performrealmevaluation" aoid="PerformRealmEvaluation">
+			<h1>PerformRealmEvaluation ( _x_, _evalRealm_ )</h1>
```
Jack-Works, line 122:
```suggestion
			<h1>PerformRealmEvaluation ( _sourceText_, _evalRealm_ )</h1>
```

spec.html
```diff
@@ -32,8 +32,118 @@ <h1>Well-known intrinsic objects</h1>
...
+			<emu-alg>
+				1. Assert: Type(_x_) is String.
+				1. Assert: _evalRealm_ is a Realm Record.
+				1. Let _hostDefined_ be _evalRealm_.[[HostDefined]].
+				1. Let _s_ be ParseScript(_x_, _evalRealm_, _hostDefined_).
+				1. If _s_ is a List of errors, then
+					1. Perform HostReportErrors(_s_).
```
Jack-Works, line 129: I don't find the definition of `HostReportErrors` in this file nor in https://tc39.es/ecma262/

spec.html
```diff
@@ -32,8 +32,118 @@ <h1>Well-known intrinsic objects</h1>
...
+					1. Return NormalCompletion(*undefined*).
```
Jack-Works, line 130: Why parse error will return a normal completion?

spec.html
```diff
@@ -32,8 +32,118 @@ <h1>Well-known intrinsic objects</h1>
...
+	<emu-clause id="sec-realm-abstracts">
+		<h1>Realm Abstract Operations</h1>
+
+		<emu-clause id="sec-performrealmevaluation" aoid="PerformRealmEvaluation">
+			<h1>PerformRealmEvaluation ( _x_, _evalRealm_ )</h1>
+			<emu-alg>
+				1. Assert: Type(_x_) is String.
+				1. Assert: _evalRealm_ is a Realm Record.
+				1. Let _hostDefined_ be _evalRealm_.[[HostDefined]].
+				1. Let _s_ be ParseScript(_x_, _evalRealm_, _hostDefined_).
```
Jack-Works, line 127: Should we do `HostEnsureCanCompileStrings` before this?

spec.html
```diff
@@ -92,19 +202,39 @@ <h1>Properties of the Realm Prototype Object</h1>
...
+				1. Return ? GetWrappedValue(_realm_, _value_).
+			</emu-alg>
+
+			<emu-note>
+				Extensible web: This is the dynamic equivalent of a &lt;script&gt; in HTML.
+			</emu-note>
+		</emu-clause>
+
+		<emu-clause id="sec-realm.prototype.importBinding">
+			<h1>Realm.prototype.importBinding ( _specifier_, _bindingName_ )</h1>
```
Jack-Works, line 225:
```suggestion
			<h1>Realm.prototype.importBinding ( _specifier_, _exportName_ )</h1>
```

spec.html
Jack-Works, line 322: Does it have the live binding?

spec.html
Jack-Works, line 330: How is it going to happen? I guess the export is static

spec.html
```diff
@@ -32,8 +32,118 @@ <h1>Well-known intrinsic objects</h1>
...
+		1. Let _target_ be _F_.[[WrappedTargetFunction]].
+		1. Assert: IsCallable(_target_) is *true*.
+		1. Let _targetRealm_ be ? GetFunctionRealm(_target_).
+		1. Let _callerRealm_ be ? GetFunctionRealm(_F_).
+		1. For each element _key_ of _argumentsList_, do
+			1. Let o be _argumentsList_[_key_].
+			1. Set _argumentsList_[_key_] to ? GetWrappedValue(_targetRealm_, _o_).
+		1. Let _wrappedThisArgument_ to ? GetWrappedValue(_targetRealm_, _thisArgument_).
+		1. Let _value_ be ? Call(_target_, _wrappedThisArgument_, _argumentsList_).
+		1. Return ? GetWrappedValue(_callerRealm_, _value_).
```
Jack-Works, line 95: A poor man's membrane. I need to code in this style to access objects in the other realm.

```js
const argParam = { a: 1 }
const returnValueGetter = wrappedFunction(name => argParam[name])
const x = returnValueGetter("x")
```

Jack-Works (changes requested):
This change looks like a "poor man's membrane" to me. I strongly suggest we should keep the goal of the membrane in mind though we can ship this `Function or primitive only` membrane for now. (which means we should not add anything that might conflict with the membrane if we need to add it in the future).

And I don't think removing the original unwrapped version is a good idea. This limits the full power of Realms API. Please also keep the old version, and make the Isolated version as an opt-in advance feature.

spec.html
```diff
@@ -32,8 +32,118 @@ <h1>Well-known intrinsic objects</h1>
...
+	<emu-alg>
+		1. Let _target_ be _F_.[[WrappedTargetFunction]].
+		1. Assert: IsCallable(_target_) is *true*.
+		1. Let _targetRealm_ be ? GetFunctionRealm(_target_).
+		1. Let _callerRealm_ be ? GetFunctionRealm(_F_).
+		1. For each element _key_ of _argumentsList_, do
+			1. Let o be _argumentsList_[_key_].
+			1. Set _argumentsList_[_key_] to ? GetWrappedValue(_targetRealm_, _o_).
+		1. Let _wrappedThisArgument_ to ? GetWrappedValue(_targetRealm_, _thisArgument_).
+		1. Let _value_ be ? Call(_target_, _wrappedThisArgument_, _argumentsList_).
```
Jack-Works, line 94: `? Call` will leak error objects from another realm.

```js
// other.js
export function x() { null.a }
// main.js
try {
     new Realm().importBinding("./other.js", "x")()
} catch(e) {
    e instanceof TypeError // but in another realm
}
```

spec.html
```diff
@@ -32,8 +32,118 @@ <h1>Well-known intrinsic objects</h1>
...
+			</emu-alg>
+		</emu-clause>
+
+		<emu-clause id="sec-getwrappedvalue" aoid="GetWrappedValue">
+			<h1>GetWrappedValue ( _callerRealm_, _value_ )</h1>
+			<emu-alg>
+				1. Assert: _callerRealm_ is a Realm Record.
+				1. If Type(_value_) is Object, then
+				  1. If IsCallable(_value_) is *false*, throw a TypeError exception.
+				  1. Return ? WrappedFunctionCreate(_callerRealm_, _value_).
```
Jack-Works, line 141: It will return a different function even the `_value_`-`_callerRealm_` pair is the same.

leobalter:
Thanks for the review @Jack-Works.

> And I don't think removing the original unwrapped version is a good idea. This limits the full power of Realms API. Please also keep the old version, and make the Isolated version as an opt-in advance feature.

As much as I want to have full power of Realms API, the goal is to land something we can work with. The isolated realms is good enough and provides tools that enables the goals of this proposal, without being challenged by the invariants once presented to the champions and TC39 that historically prevented advancement to Stage 3.

I'll cheer up if someone wants to take the previous Realms API forward, but as a champion of this proposal, I'd rather stick with the isolated Realms for now and explore membranes over the given API.

To be honest, I like the path we are going towards avoiding identity discontinuity. We are setting a good exploration path towards injecting code in the realms and that's the field I'd like for us to explore more next.

---

Please refrain the goal of this PR is just to update the current spec draft of the proposal. I'd like to ask to avoid using this thread to discuss the general directions of this proposal. At the same time, I'm requesting we stick with all the TC39 conversation we already had about this and what lead us to where we are right now.

spec.html
leobalter: I have no idea what you mean by this code, but I'm pretty sure you can use the current API to get wrapped functions and primitive values. We hope that complex structures such as argParam can eventually be transferred as Records and Tuples, but that depends on the respective proposal to land.

Also, this is a discussion already existing in the other issues. Bringing this discussion here unnecessarily expands the complexity of integrating this PR into the main branch. Let's use the proper channels.

spec.html
caridy: thanks for pointing this out.

spec.html
caridy: Probably from old versions of 262, this was a revert from old versions of this specs anyways. I will fix all that.

spec.html
caridy: yes, missing.

spec.html
caridy: that is by design as explained in https://github.com/tc39/proposal-realms/issues/289, this is not a membrane, it does not do identity reconciliation. Every time it sees a callable coming thru the boundary, a new wrapped function is created.

spec.html
caridy: fixed!

spec.html
caridy: that's correct, if you want the new value in the future, you have to invoke `importBinding` again from the realm. you're not getting back an exotic namespace object with bindings are properties, you are getting back a value.

spec.html
caridy: that's correct, it is not a declaration of a binding, it is just invoking an API to resolve to a value.

spec.html
caridy: fixed.

caridy:
@leobalter @Jack-Works I have fixed the issues you have pointed out, this is ready for another pass.

leobalter (approved):
From here we might have just details here and there. I want to merge this for now and leave any fixes to follow up PRs.

spec.html
Jack-Works: If we don't keep identity reconciliation, we will never be able to make it after we ship it cause it's breaking change. That will block the possibility to the membrane

spec.html
Jack-Works: Why not? I think the exotic namespace object with bindings are good.

spec.html
caridy: No, that's not true, you can implement identity preserving membrane on top of this.

mhofman: It would be the responsibility of the membrane or library sitting on top of this API to avoid recreating wrappers.

Given that symbols are unique and comparable through round trips, wrapper functions provide indirect cross realms references, and WeakRefs are available, it should be possible to build a full non-leaky proxy based membrane.  I believe the next step is to validate this by building such a membrane on top of this proposed API. See https://github.com/leobalter/realms-polyfill/issues/13

mhofman: One question however is if the `GetWrappedValue` operation should automatically unwrap a wrapped function if its `[[Realm]]` matches `callerRealm`. Right now it would double wrap it.

leobalter: I'm opening an issue to discuss if we should unwrap or double wrap it.

I don't see much value or the goal in sending the same function back and forth. If the goal is to get a unique identity, symbols should resolve this trick in a much more lightweight manner. 

Adding an extra check for every function to see if it can be unwrapped might end with some cost, but at the same time, it cuts a chain of references here and there.

PR #292, opened 2021-03-27
Meta: Update Explainer for Isolated Realms

leobalter:


explainer.md
```diff
@@ -288,14 +307,13 @@ We still want things to interact with the DOM without spending any excessive amo
...
 
-```js
-import virtualDocument from 'virtual-document';
-
+```javascript
 const realm = new Realm();
 
-realm.globalThis.document = virtualDocument;
+const initVirtualDocument = await realm.importBinding('virtual-document', 'init');
+await realm.import('./publisher-amin.js', 'symbolId');
```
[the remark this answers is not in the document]
leobalter: Yes, I want to explore allowing no bindings, but I'm also avoiding justification over the illustrated examples only. This `symbolId` is here just for the requirement of importing a binding, it smells.

explainer.md
```diff
@@ -316,52 +334,70 @@ The Realms API provides a single API for this virtualization in both browsers an
...
-    const someRealmIntrinsicsNeededForWrappers = extractIntrinsicsFromGlobal(realmGlobalThis);
-    Object.defineProperties({
-         document: createFakeDocumentDescriptor(someRealmIntrinsicsNeededForWrappers),
-         Element: createFakeElementDescriptor(someRealmIntrinsicsNeededForWrappers),
-         Node: ...
-         ... // all necessary DOM related globals should be defined here
-    });
+```javascript
+export default function() {
+  const someRealmIntrinsicsNeededForWrappers = extractIntrinsicsFromGlobal(customGlobalThis);
```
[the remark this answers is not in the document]
leobalter: I need to review this example with @caridy, it seems like the approach would be different now and there is more to be changed here.

explainer.md
```diff
@@ -316,52 +334,70 @@ The Realms API provides a single API for this virtualization in both browsers an
...
 
-await installFakeDOM(realm.globalThis);
+const installFakeDOM = await realm.importBinding('./fakedom.js', 'default');
 
-// Custom properties are now added to the Realm such as
-realm.globalThis.document;
-realm.globalThis.top;
-realm.globalThis.location;
+// Custom properties can be added to the Realm
+// The resolution callback is used as the current API can't wrap promises
```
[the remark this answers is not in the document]
leobalter: yes, I wrote this while modifying the example. It makes no sense so I removed it.

explainer.md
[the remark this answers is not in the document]
caridy: yeah, let's chat about it

explainer.md
```diff
@@ -153,22 +158,23 @@ Instances of Realm Objects and their Global Objects have their lifeline to their
 
 ### <a name='Evaluation'></a>Evaluation
 
-The Realms API does not introduce a new way to evaluate code, it is subject to the existing evaluation mechanisms such as the [Content-Security-Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy).
+The Realms API does not introduce a new way to evaluate code. Any code is subject to the existing evaluation mechanisms such as the [Content-Security-Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy).
```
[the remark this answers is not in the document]
leobalter: This new mechanism is basically indirect eval, or just mirrors indirect eval, with the restriction that `evaluate` won't accept non-string values as argument and won't be able to return Non Callable Objects. 

I'm tweaking this out for clarification but I believe one could still debate saying this isn't a new evaluation mechanism.

explainer.md
```diff
@@ -597,56 +562,18 @@ iframeWindow.top;
...
-2. For each property of the Global Object specified in clause 18, do
-...
-
-18 The Global Object
-
-...
-- may have host defined properties in addition to the properties defined in this specification. This may include a property whose value is the global object itself.
-
-```
+Yes! It only creates a new copy of the built-ins from ECMAScript.
```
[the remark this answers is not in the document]
leobalter: I haven't discussed that yet but I believe this new API model leads me to insist more in just providing a copy of the intrinsics. I hope this API resolves that matter.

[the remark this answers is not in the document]
leobalter: 6220f8e

README.md
```diff
@@ -50,6 +50,7 @@ See some examples [in the Explainer file](explainer.md).
 
 ## <a name='History'></a>History
 
+* we moved on from the exposed globalThis model to a lean isolated realms API (see #289 and #291)
```
caridy, line 53: isolated realms => Realm with Callable Boundary or something like that...

we chatted about this with Shu, and I believe we should avoid using the term isolated realm all together.

caridy: Another alternative is also "Callable Realms"

explainer.md
```diff
@@ -107,7 +112,7 @@ These programs must currently contend for the global shared resources, specifica
 
 Attempting to solve these problems with existing DOM APIs will require to implement an asynchronous communication protocol, which is often a deal-breaker for many use cases. It usually just adds complexity for cases where a same-process Realm is sufficient. It's also very important that values can be immediately shared. Other communications require data to be serialized before it's sent back and forth.
 
-__The primary goal of this proposal is to provide a proper mechanism to control the execution of a program, providing a new global object, a new set of intrinsics, no default access to the incubator realm's global object, a separate module graph and synchronous communication with the incubator realm__.
+__The primary goal of this proposal is to provide a proper mechanism to control the execution of a program, providing a new global object, a new set of intrinsics, no immediate access to objects cross-realms, a separate module graph and synchronous communication between both realms, while not forcing fingerprints in the constructed realm__.
```
caridy, line 115:
```suggestion
__The primary goal of this proposal is to provide a proper mechanism to control the execution of a program, providing a new global object, a new set of intrinsics, no access to objects cross-realms, a separate module graph and synchronous communication between both realms, while not forcing fingerprints in the constructed realm__.
```

caridy: I also continue having a hard time understanding the finderprints issues with the previous proposal. can you elaborate more?

littledan: I'm skeptical that this proposal changes the fingerprinting surface. I think that it'd be best to remove this claim.

explainer.md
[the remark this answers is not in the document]
littledan: I agree with you; I was making a very narrow comment about the wording. This is very much analogous to things that CSP already controls.

README.md
littledan: I find the names that @caridy is suggesting confusing, though avoiding the word "isolated" could be useful to avoid other kinds of confusion. Maybe we could avoid giving this version a name for now, until we're really confident in one, since it's awkward to change names later.

explainer.md
```diff
@@ -268,18 +270,35 @@ A big codebase tend to evolve slowly and soon becomes legacy code. Old code vs n
 
 Modifying code to resolve a conflict (e.g.: global variables) is non-trivial, specially in big codebases.
 
-The Realms API can provide a _lightweight_ mechanism to preserve the integrity of the intrinsics.0 Therefore, it could isolate libraries or logical pieces of the codebase per Realm.
+The Realms API can provide a _lightweight_ mechanism to preserve the integrity of the intrinsics. Therefore, it could isolate libraries or logical pieces of the codebase per Realm.
 
 ### <a name='Templatelibraries'></a>Template libraries
```
littledan, line 275: I still think that this example should be removed. It will likely initially be too expensive to instantiate this many Realms; we don't want to be telling developers that this is expected to be cheap.

explainer.md
```diff
@@ -404,178 +437,110 @@ There are many examples like this for the web: Google Sheets, Figma's plugins, o
...
-let a2 = globalTwo.eval('[1,2,3]');
-a1.prototype === a2.prototype; // yield false
-a1 instanceof globalTwo.Array; // yield false
-a2 instanceof globalOne.Array; // yield false
-```
+This code is **not** possible with the Realms API! Non-primitive values are not transfered cross-realms using the Realms API.
 
 ## <a name='Example:NodesvmobjectsvsRealms'></a>Example: Node's vm objects vs Realms
 
 If you're using node's `vm` module today to "evaluate" javascript code in a different realm, you can replace it with a new Realm, e.g.:
```
littledan, line 495: This is no longer quite true, given the restrictions. You probably want to explain that.

explainer.md
leobalter: > I also continue having a hard time understanding the finderprints issues with the previous proposal. can you elaborate more?

That's just about identifying identity from other realms through objects that would then be accessed by `globalThis`. Of course any framework would handle that, and that's not any new complication (iframe allows that already).

I'll remove this part.

explainer.md
leobalter: I can remove, but it doesn't require many Realms. You could freeze a realm to operate the templates inside.

explainer.md
leobalter: 7b88d54

explainer.md
leobalter: 246e41b

It's not much but it provides some honest information. WDYT, @littledan?

leobalter:
@caridy @littledan I believe I addressed the latest feedback. PTAL

leobalter:
I'll be glad to address any new feedback in a follow up PR. I'm merging this now to avoid confusion before the TC39 meeting.

PR #301, opened 2021-05-04
Normative: Clarify text about host added properties

leobalter:
Ref #284

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -371,7 +371,7 @@ location: https://tc39.es/proposal-realms/
 					This specification does not recommend any specific addition.
 					In the Web embedding, HTML and WebIDL will specify which
 					interfaces are included. The Web Platform and Web-like
-					environments may decide to include `setTimeout`, `EventTarget`,
+					environments may decide to include `EventTarget`,
 					`atob`, `TextEncoder`, `URL`, etc. while at the same time not
 					including `HTMLElement`, `console`, `localStorage`, `fetch`, etc..
 				</p>
--- a/spec.html
+++ b/spec.html
@@ -357,10 +357,25 @@ location: https://tc39.es/proposal-realms/
 				operation used to inform the host of any newly created realms from
 				the Realm constructor. Its return value is not used, though it may
 				throw an exception. The idea of this hook is to initialize host
-				data structures related to the Realm, e.g., for module loading. It
-				is not expected that this hook would add properties to the Realm's
-				global object.
+				data structures related to the Realm, e.g., for module loading.
 			</p>
+			<p>
+				The host may use this hook to add properties to the Realm's global
+				object. Those properties must each be configurable to provide
+				platform capabilities with no authority to cause side effects such
+				as I/O or mutation of values that are shared across different
+				realms within the same host environment.
+			</p>
+			<emu-note>
+				<p>
+					EDITOR'S NOTE: This specification does not recommend any specific
+					addition. In the Web embedding, HTML and WebIDL will specify which
+					interfaces are included. The Web Platform and Web-like
+					environments may decide to include `setTimeout`, `EventTarget`,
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -357,10 +357,25 @@ <h1>Runtime Semantics: HostInitializeSyntheticRealm ( _realm_ )</h1>
...
+				platform capabilities with no authority to cause side effects such
+				as I/O or mutation of values that are shared across different
+				realms within the same host environment.
+			</p>
+			<emu-note>
+				<p>
+					This specification does not recommend any specific addition.
+					In the Web embedding, HTML and WebIDL will specify which
+					interfaces are included. The Web Platform and Web-like
+					environments may decide to include `setTimeout`, `EventTarget`,
```
Jack-Works, line 374: AFAIK setTimeout is IO. And as I explained in #284 I strongly against the host being able to add things into the realm without being able to have a clean version. This also hurts the API to be cross-platform safe.

Jack-Works (changes requested):

spec.html
leobalter: This is just an editorial note and we can remove the example. 

Anyway, please keep the discussion to the respective issue.

littledan (approved):
This PR is a clean way to describe the changes due to the feedback in https://github.com/tc39/proposal-realms/issues/284#issuecomment-728964261 . We will need to follow up with details in the HTML side at some point, but I don't see why that should block Stage 3 (though it will block shipping in web browsers).

spec.html
```diff
@@ -357,10 +357,25 @@ <h1>Runtime Semantics: HostInitializeSyntheticRealm ( _realm_ )</h1>
...
-				is not expected that this hook would add properties to the Realm's
-				global object.
+				data structures related to the Realm, e.g., for module loading.
 			</p>
+			<p>
+				The host may use this hook to add properties to the Realm's global
+				object. Those properties must each be configurable to provide
+				platform capabilities with no authority to cause side effects such
+				as I/O or mutation of values that are shared across different
+				realms within the same host environment.
```
Jack-Works, line 367: > with no authority to cause side effects such as I/O or mutation of values that are shared across different realms

This is not enough for the SES requirement.

Imagine the host provide a `counter { [[Get]]: () => #counter, [[Set]]: (val) => #counter = val }` property on the Realm's global object, this provides a shared global state _within_ the Realm (not shared across different realms). It provides a communication channel in the different compartments within the same ream.

Jack-Works (changes requested):

ljharb:
Something known that would block shipping in web browsers should of course block stage 3 - since the point of stage 3 is to ship it in web browsers.

leobalter:
My guess is that a stage 3 would facilitate implementation without holding on what properties are added through HTML. This clears path for experimenting in this integration, while providing a fairly stable contract of a spec candidate from the ECMAScript side.

littledan:
> Something known that would block shipping in web browsers should of course block stage 3 - since the point of stage 3 is to ship it in web browsers.

I don't think this is how we've been doing things. We have frequently continued to iterate on HTML integration PRs after Stage 3 in recent proposals. test262 is also usually required for browsers to ship, but it's not a Stage 3 requirement. I think we should leave it up to the HTML folks whether they feel like things are complete enough on their end to go to Stage 3.

spec.html
littledan: I guess I read the text to cover that case--I see the code sample as a value shared across realms, at a high level. I guess we just need to iterate on the wording so that it's maximally clear.

caridy (approved):

spec.html
Jack-Works: If `#counter` is a host-defined pre-realm internal state, it will be a mutable state within the Realm and passes the current spec text but violate the SES requirement.

PR #303, opened 2021-05-05
Remove unnecessary IfAbruptRejectPromise from PerformRealmImportValue (fixes #302)

arai-a:


spec.html
```diff
@@ -182,7 +182,6 @@ <h1>PerformRealmImportValue ( _specifierString_, _exportNameString_, _callerReal
 				1. Assert: _callerRealm_ is a Realm Record.
 				1. Assert: _evalRealm_ is a Realm Record.
 				1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
-				1. IfAbruptRejectPromise(_specifierString_, _promiseCapability_).
```
caridy, line 185: good catch! I copied some of this from https://tc39.es/ecma262/#sec-import-call-runtime-semantics-evaluation, and never clean up that step.

caridy (approved):

PR #309, opened 2021-05-20
Spec fixes, applies review feedback

leobalter:
Ref https://github.com/tc39/proposal-realms/issues/304#issuecomment-842736704

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -221,7 +221,7 @@ location: https://tc39.es/proposal-realms/
 
 		<emu-clause id="sec-realm">
 			<h1>Realm ()</h1>
-			<p>When the `Realm` function is called with no arguments, the following steps are taken:</p>
+			<p>When the `Realm` function is called, the following steps are taken:</p>
 			<emu-alg>
 				1. If NewTarget is *undefined*, throw a *TypeError* exception.
 				1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, "%Realm.prototype%", « [[Realm]], [[ExecutionContext]] »).
--- a/spec.html
+++ b/spec.html
@@ -142,9 +142,9 @@ location: https://tc39.es/proposal-realms/
 					1. If _script_ is a List of errors, throw a *SyntaxError* exception.
 					1. If _script_ Contains |ScriptBody| is *false*, return *undefined*.
 					1. Let _body_ be the |ScriptBody| of _script_.
-					1. If _body_ Contains |NewTarget|, throw a *SyntaxError* exception.
-					1. If _body_ Contains |SuperProperty|, throw a *SyntaxError* exception.
-					1. If _body_ Contains |SuperCall|, throw a *SyntaxError* exception.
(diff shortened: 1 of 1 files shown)


codehag (approved):
This covers the always false variables and the realms wording from my review. stamped from my side

spec.html
```diff
@@ -136,17 +137,14 @@ <h1>PerformRealmEval ( _sourceText_, _callerRealm_, _evalRealm_ )</h1>
...
-				1. Let _inDerivedConstructor_ be *false*.
 				1. Perform the following substeps in an implementation-defined order, possibly interleaving parsing and error detection:
 					1. Let _script_ be ParseText(! StringToCodePoints(_sourceText_), |Script|).
 					1. If _script_ is a List of errors, throw a *SyntaxError* exception.
 					1. If _script_ Contains |ScriptBody| is *false*, return *undefined*.
 					1. Let _body_ be the |ScriptBody| of _script_.
-					1. If _inFunction_ is *false*, and _body_ Contains |NewTarget|, throw a *SyntaxError* exception.
-					1. If _inMethod_ is *false*, and _body_ Contains |SuperProperty|, throw a *SyntaxError* exception.
-					1. If _inDerivedConstructor_ is *false*, and _body_ Contains |SuperCall|, throw a *SyntaxError* exception.
+					1. If _body_ Contains |NewTarget|, throw a *SyntaxError* exception.
```
syg, line 145: Nit: Here and below, it should `_body_ Contains |Foo| is *true*`

syg (approved):
lgtm, thanks

spec.html
leobalter: 4e00347

spec.html
```diff
@@ -136,17 +137,14 @@ <h1>PerformRealmEval ( _sourceText_, _callerRealm_, _evalRealm_ )</h1>
 				1. Assert: _callerRealm_ is a Realm Record.
 				1. Assert: _evalRealm_ is a Realm Record.
 				1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _evalRealm_).
-				1. Let _inFunction_ be *false*.
```
caridy, line 139: this form was copied from https://tc39.es/ecma262/#sec-performeval. My bad, because the flag for direct eval is not longer needed, therefore these lets declarations are not needed anymore. Good catch!

spec.html
```diff
@@ -93,15 +93,16 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
 			1. Assert: IsCallable(_target_) is *true*.
 			1. Let _targetRealm_ be ? GetFunctionRealm(_target_).
 			1. Let _callerRealm_ be ? GetFunctionRealm(_F_).
-			1. For each element _key_ of _argumentsList_, do
-				1. Let _o_ be _argumentsList_[_key_].
-				1. Set _argumentsList_[_key_] to ? GetWrappedValue(_targetRealm_, _o_).
+			1. Let _wrappedArgs_ be a new empty List.
```
caridy, line 96: this was me being lazy! jajaja... good catch!

caridy (approved):

PR #311, opened 2021-05-20
Normative: Remove HostImportModuleBindingDynamically hook

littledan:
Instead, its single usage is replaced by a usage of HostImportModuleDynamically

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -181,19 +181,29 @@ location: https://tc39.es/proposal-realms/
 				1. Assert: Type(_exportNameString_) is String.
 				1. Assert: _callerRealm_ is a Realm Record.
 				1. Assert: _evalRealm_ is a Realm Record.
-				1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+				1. Let _innerCapability_ be ! NewPromiseCapability(%Promise%).
 				1. Let _runningContext_ be the running execution context.
 				1. If _runningContext_ is not already suspended, suspend _runningContext_.
 				1. Push _evalContext_ onto the execution context stack; _evalContext_ is now the running execution context.
-				1. Perform ! HostImportModuleBindingDynamically(_callerRealm_, _evalRealm_, _specifierString_, _exportNameString_, _promiseCapability_).
+				1. Perform ! HostImportModuleDynamically(*null*, _specifierString_, _innerCapability_).
 				1. Suspend _evalContext_ and remove it from the execution context stack.
 				1. Resume the context that is now on the top of the execution context stack as the running execution context.
-				1. Return _promiseCapability_.[[Promise]].
+				1. Let _steps_ be the steps of an ExportGetter function as described below.
+				1. Let _onFulfilled_ be ! CreateBuiltinFunction(_steps_, 1, "", « [[ExportNameString]] », _callerRealm_).
+				1. Set _onFulfilled_.[[ExportNameString]] to _exportNameString_.
+				1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+				1. Return ! PerformPromiseThen(_innerCapability_.[[Promise]], _onFulfilled_, _callerRealm_.[[Intrinsics]].[[%ThrowTypeError%]], _promiseCapability_).
+				1. Return _promiseCapability.[[Promise]].
 			</emu-alg>
 
-			<emu-note type=editor>
-				The Realm's Execution Context is passed as a parameter _evalContext_. An alternative phrasing would be to create an extra execution context around the call to HostImportModuleBindingDynamically in order to execute within the appropriate Realm--the host can reference the current Realm to figure out how the import should work, similar to PerformRealmEval, which could be revisited when merging into the main specification.
-			</emu-note>
+			<p>An ExportGetter function is an anonymous built-in function with a [[ExportNameString]] internal slot. When an ExportGetter function is called with argument _exports_, it performs the following steps:</p>
+			<emu-alg>
+				1. Let _f_ be the active function object.
+				1. Let _string_ be _f_.[[ExportNameString]].
+				1. Let _realm_ be _f_.[[Realm]].
+				1. Let _value_ be ? Get(_exports_, _string_).
+				1. Return ? GetWrappedValue(_realm_, _value_).
+			</emu-alg>
 		</emu-clause>
 
 		<emu-clause id="sec-getwrappedvalue" aoid="GetWrappedValue">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -181,19 +181,29 @@ <h1>RealmImportValue ( _specifierString_, _exportNameString_, _callerRealm_, _ev
 				1. Assert: Type(_exportNameString_) is String.
 				1. Assert: _callerRealm_ is a Realm Record.
 				1. Assert: _evalRealm_ is a Realm Record.
-				1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+				1. Let _innerCapability_ be ! NewPromiseCapability(%Promise%).
 				1. Let _runningContext_ be the running execution context.
 				1. If _runningContext_ is not already suspended, suspend _runningContext_.
 				1. Push _evalContext_ onto the execution context stack; _evalContext_ is now the running execution context.
-				1. Perform ! HostImportModuleBindingDynamically(_callerRealm_, _evalRealm_, _specifierString_, _exportNameString_, _promiseCapability_).
+				1. Perform ! HostImportModuleDynamically(*null*, _specifierString_, _innerCapability_).
```
leobalter, line 188: I really like this `HostImportModuleDynamically` goes well with the running execution context shifting to the target realm and restored after.

spec.html
```diff
@@ -181,19 +181,29 @@ <h1>RealmImportValue ( _specifierString_, _exportNameString_, _callerRealm_, _ev
...
 				1. Suspend _evalContext_ and remove it from the execution context stack.
 				1. Resume the context that is now on the top of the execution context stack as the running execution context.
-				1. Return _promiseCapability_.[[Promise]].
+				1. Let _steps_ be the steps of an ExportGetter function as described below.
+				1. Let _onFulfilled_ be ! CreateBuiltinFunction(_steps_, 1, "", « [[ExportNameString]] », _callerRealm_).
+				1. Set _onFulfilled_.[[ExportNameString]] to _exportNameString_.
+				1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+				1. Return ! PerformPromiseThen(_innerCapability_.[[Promise]], _onFulfilled_, _callerRealm_.[[Intrinsics]].[[%ThrowTypeError%]], _promiseCapability_).
+				1. Return _promiseCapability.[[Promise]].
```
leobalter, line 196: double returns

spec.html
```diff
@@ -181,19 +181,29 @@ <h1>RealmImportValue ( _specifierString_, _exportNameString_, _callerRealm_, _ev
...
 
-			<emu-note type=editor>
-				The Realm's Execution Context is passed as a parameter _evalContext_. An alternative phrasing would be to create an extra execution context around the call to HostImportModuleBindingDynamically in order to execute within the appropriate Realm--the host can reference the current Realm to figure out how the import should work, similar to PerformRealmEval, which could be revisited when merging into the main specification.
-			</emu-note>
+			<p>An ExportGetter function is an anonymous built-in function with a [[ExportNameString]] internal slot. When an ExportGetter function is called with argument _exports_, it performs the following steps:</p>
+			<emu-alg>
+				1. Let _f_ be the active function object.
+				1. Let _string_ be _f_.[[ExportNameString]].
+				1. Let _realm_ be _f_.[[Realm]].
+				1. Let _value_ be ? Get(_exports_, _string_).
```
leobalter, line 204: I believe this should fail if exports has not a matching own property

leobalter (changes requested):
This is great! I left some small comments, thank you!

spec.html
```diff
@@ -181,19 +181,29 @@ <h1>RealmImportValue ( _specifierString_, _exportNameString_, _callerRealm_, _ev
...
-				1. Perform ! HostImportModuleBindingDynamically(_callerRealm_, _evalRealm_, _specifierString_, _exportNameString_, _promiseCapability_).
+				1. Perform ! HostImportModuleDynamically(*null*, _specifierString_, _innerCapability_).
 				1. Suspend _evalContext_ and remove it from the execution context stack.
 				1. Resume the context that is now on the top of the execution context stack as the running execution context.
-				1. Return _promiseCapability_.[[Promise]].
+				1. Let _steps_ be the steps of an ExportGetter function as described below.
+				1. Let _onFulfilled_ be ! CreateBuiltinFunction(_steps_, 1, "", « [[ExportNameString]] », _callerRealm_).
+				1. Set _onFulfilled_.[[ExportNameString]] to _exportNameString_.
+				1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+				1. Return ! PerformPromiseThen(_innerCapability_.[[Promise]], _onFulfilled_, _callerRealm_.[[Intrinsics]].[[%ThrowTypeError%]], _promiseCapability_).
```
caridy, line 195: I'm not sure it should be a type error, in the previous version I was more hand waving (intentionally). When you fail to wrap, the error is not a type error either, isn't it?

caridy (approved):
This is great stuff, I tried to do this before but could not find a way to implement it in spec language :(, now I know!

The only question is really about the type of error to be used.

spec.html
leobalter:
```js
import('bla').then(() => console.log('noop')).catch(console.log);
// TypeError: Failed to resolve module specifier 'bla'
//    at <anonymous>:1:1
```

In general this would also wrap import errors into TypeError (if module 'bla' exists and throws an error on evaluation).

spec.html
guybedford, line 188: I'd suggest extending `HostImportModuleDynamically` to accept a realm argument:

```suggestion
				1. Perform ! HostImportModuleDynamically(*null*, _specifierString_, _evalRealm_, _innerCapability_).
```

Then to update the spec text for `HostImportModuleDynamically` to note that all module records should have their realm set to this realm. That seems a nice generalization as well.

spec.html
```diff
@@ -181,19 +181,29 @@ <h1>RealmImportValue ( _specifierString_, _exportNameString_, _callerRealm_, _ev
 				1. Assert: Type(_exportNameString_) is String.
 				1. Assert: _callerRealm_ is a Realm Record.
 				1. Assert: _evalRealm_ is a Realm Record.
-				1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+				1. Let _innerCapability_ be ! NewPromiseCapability(%Promise%).
 				1. Let _runningContext_ be the running execution context.
 				1. If _runningContext_ is not already suspended, suspend _runningContext_.
 				1. Push _evalContext_ onto the execution context stack; _evalContext_ is now the running execution context.
```
guybedford, line 187: Handling the running context shouldn't be necessary here at all as it is handled per module at the execution algorithm level.

guybedford (approved):
This looks great - much clearer to see what is going on.

If there are further restrictions / invariants being applied to the module record map interactions, it could be nice to include those details in a spec note.

spec.html
leobalter: That's great, but involves circling back with the ECMA-262 editors. I'm moving this to a new issue to avoid losing track of it.

PR #313, opened 2021-05-21
Normative: Require property to exist from the imported module namespace

leobalter:
Ref https://github.com/tc39/proposal-realms/pull/311#discussion_r636479941

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -201,10 +201,11 @@ location: https://tc39.es/proposal-realms/
 				1. Let _f_ be the active function object.
 				1. Let _string_ be _f_.[[ExportNameString]].
 				1. Assert: Type(_string_) is String.
-				1. Let _Desc_ be ? _exports_.[[GetOwnProperty]](_string_).
-				1. If _Desc_ is *undefined*, throw a *TypeError* exception.
+				1. Let _hasOwn_ be ? HasOwnProperty(_exports_, _string_).
+				1. If _hasOwn_ is *false*, throw a *TypeError* exception.
+				1. Let _value_ be ? Get(_exports_, _string_).
 				1. Let _realm_ be _f_.[[Realm]].
-				1. Return ? GetWrappedValue(_realm_, _Desc_.[[Value]]).
+				1. Return ? GetWrappedValue(_realm_, _value_).
 			</emu-alg>
 		</emu-clause>
 
--- a/spec.html
+++ b/spec.html
@@ -197,11 +197,14 @@ location: https://tc39.es/proposal-realms/
 
 			<p>An ExportGetter function is an anonymous built-in function with a [[ExportNameString]] internal slot. When an ExportGetter function is called with argument _exports_, it performs the following steps:</p>
 			<emu-alg>
+				1. Assert: _exports_ is a module namespace exotic object.
 				1. Let _f_ be the active function object.
 				1. Let _string_ be _f_.[[ExportNameString]].
+				1. Assert: Type(_string_) is String.
+				1. Let _Desc_ be ? _exports_.[[GetOwnProperty]](_string_).
+				1. If _Desc_ is *undefined*, throw a *TypeError* exception.
 				1. Let _realm_ be _f_.[[Realm]].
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -197,11 +197,14 @@ <h1>RealmImportValue ( _specifierString_, _exportNameString_, _callerRealm_, _ev
 
 			<p>An ExportGetter function is an anonymous built-in function with a [[ExportNameString]] internal slot. When an ExportGetter function is called with argument _exports_, it performs the following steps:</p>
 			<emu-alg>
+				1. Assert: _exports_ is a module namespace exotic object.
 				1. Let _f_ be the active function object.
 				1. Let _string_ be _f_.[[ExportNameString]].
+				1. Assert: Type(_string_) is String.
+				1. Let _Desc_ be ? _exports_.[[GetOwnProperty]](_string_).
```
ljharb, line 204: Can this use HasOwnProperty instead of directly calling the internal method?

littledan: The current spelling seems fine to me and consistent with other parts of the spec, but I would also be OK with the proposed change.

leobalter: Thanks! there might be a catch here...

In this case I'd need to check existence with HasOwnProperty, and preserve `Get` to access the exotic [`[[Get]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver). I want to avoid dynamic binding too but I might not be the case.

littledan (approved):
LGTM, being more strict here makes sense to me.

spec.html
leobalter: @ljharb I added 88c8156, but this now will trigger `[[Get]]` twice. I prefer the original change but I'm fine either way.

ljharb: oops, you're right, we wouldn't want to do two observable lookups.

The intention was to avoid (as has been strived for in the spec for many years) direct references to internal methods. However, there's no "GetOwnProperty" analog to "HasOwnProperty".

In this case though, why does it matter if it's an own property? you've already asserted it's a module namespace exotic object, which means it can only have a null prototype, which means that non-own properties *can't exist* on it.

This also means it can't be a Proxy, so altho a Get() might be observable if it's even possible to have a getter on a module namespace exotic object, a HasOwnProperty won't be (it invokes [[GetOwnProperty]] which does not actually invoke a getter).

leobalter: > In this case though, why does it matter if it's an own property?

It actually matters if the binding exists.

Today if I roll `const x = await realm.importValue('./file.js', 'bindingName');` I want it to be somewhat equivalent to `import { bindingName as x } from './file.js'`.

In this case, for the importValue, I want to make sure if `bindingName` resolves to `undefined` that is because the name exists and its value is undefined. The purpose of this PR it solely to avoid `const x = await realm.importValue('./file.js', 'bindingName');` and if `x` is undefined the reason would be ambiguous. See the exotic object [`[[Get]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver) returning undefined here if the property is not part of the exports.

ljharb: It totally matters if it exists! but what i mean is, you can Get() it, and as long as it's Type() string, or even just "not undefined", it seems like it'd be fine?

ljharb: ah, i see what you mean. in that case, yes, you'd need either HasProperty or HasOwnProperty, plus Get - or, [[GetOwnProperty]]. I prefer the former, but either sufficces.

Jack-Works: How can it be observable? Module objects has exotic and immutable prototypes IMO, perform Get on it cannot be captured by the user land 🤔

ljharb: @Jack-Works i *think* it might be possible for a module namespace object to have a getter (some host-provided module type) but i'm not sure.

leobalter: Being observable is not really the problem here. This is editorial style.

ljharb (approved):

leobalter:
cc @tc39/ecma262-editors PTAL at the 714e879 and the current version (being merged following PR approval).

PR #316, opened 2021-06-09
Normative: Relax restrictions from the HostInitializeSyntheticRealm

leobalter:


kriskowal (approved):
This is consistent with what we discussed at the SES meeting. Thanks for the follow-through.

spec.html
```diff
@@ -388,6 +385,13 @@ <h1>Runtime Semantics: HostInitializeSyntheticRealm ( _realm_ )</h1>
 					including `HTMLElement`, `console`, `localStorage`, `fetch`, etc..
 				</p>
 			</emu-note>
+			<emu-note type=editor>
+				<p>
+					The Realm constructor (<emu-xref href="#sec-realm"></emu-xref>
+					creates a new global object as an ordinary object. This
+					means all properties from the global object are deletable.
```
Jack-Works, line 392: What if an exotic object removes the property but adds it back in the future?

kriskowal:
There is a provision that within a realm the globalThis must be an ordinary
object

spec.html
caridy: @Jack-Works then the global object is not an ordinary object :)

@leobalter I thought this was implicit based on the call to create a new global obj with the values provided, which implies an ordinary object, but I'm fine with this being here.

leobalter: On [line 388](https://github.com/tc39/proposal-realms/pull/316/files#diff-181371b08d71216599b0acccbaabd03c306da6de142ea6275c2135810999805aR388) this is flagged just an editor note, meaning it's a note of this proposal draft for extra clarification.

mhofman:
> > What if an exotic object removes the property but adds it back in the future?
>
> @Jack-Works then the global object is not an ordinary object :)

Is there any invariant anywhere that says the host cannot add properties to an ordinary object, in particular `globalThis` in an "unexpected way", e.g. without any user code invoking a host function or host hook executing.

I don't think any host currently does this, but if a host mutates the global context of its own accord after the initial user code executes, it would enable a lockdown bypass.

PR #319, opened 2021-07-12
Normative: Coerce evaluate's argument to string

leobalter:
Ref #318

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -292,12 +292,15 @@ location: https://tc39.es/proposal-realms/
 			<emu-alg>
 				1. Let _O_ be *this* value.
 				1. Perform ? ValidateRealmObject(_O_).
-				1. If Type(_sourceText_) is not String, throw a *TypeError* exception.
+				1. Let _sourceStr_ be ? ToString(_sourceText_).
 				1. Let _callerRealm_ be the current Realm Record.
 				1. Let _evalRealm_ be _O_.[[Realm]].
-				1. Return ? PerformRealmEval(_sourceText_, _callerRealm_, _evalRealm_).
+				1. Return ? PerformRealmEval(_sourceStr_, _callerRealm_, _evalRealm_).
 			</emu-alg>
 
+			<emu-note type=editor>
+				The argument _sourceText_ is coerced to a string value and this behavior differs from <emu-xref href="#sec-eval-x"></emu-xref>, which has a different behavior for non-String arguments. `eval(x)` returns `x` - with no evaluation - if `x` is not a String.
+			</emu-note>
 			<emu-note type=editor>
 				Extensible web: This is the dynamic equivalent of a &lt;script&gt; in HTML.
 			</emu-note>


leobalter:
It's important to highlight @caridy's PoV from https://github.com/tc39/proposal-realms/issues/304#issuecomment-849766915 here:

>  it is that strict to avoid the confusion with eval, which has a bizarre behavior by returning the value of the first argument if it is not a string value. As for the second one, I believe it is a trying to match the semantics of dynamic import, which does not exhibit any bizarre behavior. An alternative here is to make evaluate to perform type coercion, which I'm open to, which will mean people moving from eval(value) to myRealm.evaluate(value) will not result in an error, but the behavior will not match the original behavior of eval either.

spec.html
```diff
@@ -292,12 +292,15 @@ <h1>Realm.prototype.evaluate ( _sourceText_ )</h1>
 			<emu-alg>
 				1. Let _O_ be *this* value.
 				1. Perform ? ValidateRealmObject(_O_).
-				1. If Type(_sourceText_) is not String, throw a *TypeError* exception.
+				1. Let _sourceStr_ be ? ToString(_sourceText_).
```
Jack-Works, line 295: Why not keep it throw? 👀

leobalter: This came in as a suggestion from @gibson042's spec review (#304).

As I pointed out in the comment above, @caridy shared some agreement to this change in https://github.com/tc39/proposal-realms/issues/304#issuecomment-849766915 and #318.

The high level overview for the change is to make the evaluate method consistent with `importValue` and other general methods often seen in ECMAScript. 

The main reason it throws today is to make it slightly consistent with `eval`, but the consistency is not perfect. `eval` will just return the argument if it's not a string:

```js
const obj = { toString() { return "40 + 2"; } };

eval(obj) === obj; // and not 42
```

With the Realms API, we have the current outcome:

```js
const obj = { toString() { return "40 + 2"; } };
const r = new Realm();

r.evaluate(obj); // throws TypeError
r.evaluate(new String('Math.PI')); // TypeError
```

With this PR, we could have some better usage of argument:

```js
const obj = { toString() { return "40 + 2"; } };
const r = new Realm();

r.evaluate(obj); // 42
r.evaluate(new String('Math.PI')) === Math.PI; // true, considering Math.PI is not altered in both realms
```

---

We do have precedent for things like this today, such as the `Function` constructor [coercing args and body to string](https://tc39.es/ecma262/#sec-createdynamicfunction). The approach from this PR only makes `evaluate` a bit more consistent with functionality of other methods and putting it slightly farther from `eval`. Anyway, I believe the weird behavior from `eval` on non string arguments is undesirable legacy, but part of web reality.

WDYT?

littledan: I think it'd make sense to close this PR and leave the behavior as is. `importValue` should be analogous to `import()` (which coerces) and `evaluate` should be analogous to `eval` (which returns its input when passed not a string--the natural analogue is to throw, since that behavior is actually surprising).

Jack-Works: Is it possible to accidentally enables some attack surface?

leobalter: I find validity in both sides, I don't have a strong preference and I can see a good reason either way.

littledan: I don't think this decision has any security implications. Either way we go, there's lots of similar behavior elsewhere in JS.

Jack-Works: For example, execute code accidentally on an object with `toString` that is not intended to be treated as code.

littledan: That example falls into what I mentioned above, that there's similar behavior elsewhere, so I don't think it expands the "vulnerability" meaningfully.

erights: @mikesamuel where is your proposal to use a non-string `eval` argument for a trusted-type-like branding for approved evaluation in a CSP-no-eval context? If you could answer quickly, that'd be great because we're in the midst of a tc39 meeting where this may become relevant. Thanks!

ljharb: @erights https://github.com/tc39/proposal-dynamic-code-brand-checks iirc

erights: @ljharb that's it. Thanks!

mhofman: I'm not sure I understand the benefit in coercing the source argument to a string, besides making it more similar to the `Function` constructor. And given that the callable boundary prevents access to both the other Realm's `Function.[[Construct]]` and non primitive arguments, it should be impossible to coerce to string and evaluate an object in another isolated realm that route.

spec.html
```diff
@@ -292,12 +292,15 @@ <h1>Realm.prototype.evaluate ( _sourceText_ )</h1>
...
 				1. Let _callerRealm_ be the current Realm Record.
 				1. Let _evalRealm_ be _O_.[[Realm]].
-				1. Return ? PerformRealmEval(_sourceText_, _callerRealm_, _evalRealm_).
+				1. Return ? PerformRealmEval(_sourceStr_, _callerRealm_, _evalRealm_).
 			</emu-alg>
 
+			<emu-note type=editor>
+				The argument _sourceText_ is coerced to a string value and this behavior differs from <emu-xref href="#sec-eval-x"></emu-xref>, which has a different behavior for non-String arguments. `eval(x)` returns `x` - with no evaluation - if `x` is not a String.
```
gibson042, line 302: I found [@caridy's explanation](https://github.com/tc39/proposal-realms/issues/304#issuecomment-849766915) satisfactory, and don't have strong feelings about throwing vs. coercing. But I do think a callout of the behavioral difference should remain visible in the rendered spec.

gibson042 (approved):

leobalter:
After some discussion with @caridy, we agreed to hold on and not move this PR forward for now. The reasons to keep the API as is have a good weight to also avoid such a noise before this TC39 meeting. 

I'm gonna hold on this PR and I'm pretty sure I can do a better recap to previous discussion and works such as those pointed out by @erights and @ljharb.

Thanks everyone for the quick feedback!

PR #320, opened 2021-07-12
Remove [[ExecutionContext]]

leobalter:
cc @syg @caridy @rwaldron

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -322,17 +322,10 @@ location: https://tc39.es/proposal-shadowrealm/
 					1. If _body_ Contains |SuperCall| is *true*, throw a *SyntaxError* exception.
 				1. Let _strictEval_ be ScriptIsStrict of _script_.
 				1. Let _runningContext_ be the running execution context.
-				1. Let _lexEnv_ be NewDeclarativeEnvironment(_evalRealm_.[[GlobalEnv]]).
-				1. Let _varEnv_ be _evalRealm_.[[GlobalEnv]].
-				1. If _strictEval_ is *true*, set _varEnv_ to _lexEnv_.
 				1. If _runningContext_ is not already suspended, suspend _runningContext_.
-				1. Let _evalContext_ be a new ECMAScript code execution context.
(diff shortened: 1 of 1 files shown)


littledan (approved):
This patch looks like a correct way to implement @syg's [suggestion](https://github.com/tc39/proposal-realms/issues/304#issuecomment-842736704), but I don't understand the motivation for the change. Stashing an execution context is a sort of common technique to make it easier to get into a Realm, e.g., also found in HTML's [environment settings objects](https://html.spec.whatwg.org/#environment-settings-object).

leobalter:
My only goal is unblock all the way to Stage 4, so I don't have a strong preference for this PR or otherwise.

Somehow this proposed approach makes `importValue` a bit closer to `Realm#evaluate`. 

I'm also pretty confident we're going to eventually rename the `GetNewRealmContext`.

leobalter:
Pinging @syg as I want to make sure this address his feedback.

leobalter:
pinging the @tc39/ecma262-editors here. This PR has no observable changes, but matches a request from @syg on the Stage 3 review. I'd like to confirm this is the desired direction for the eventual PR to ECMA-262.

caridy:
@ptomato is this ready for review? what is the latest here?

ptomato:
@caridy I rebased it the other day, but I don't know the context of the original request, so maybe @leobalter or @syg should have a look before merging it.

caridy:
maybe @nicolo-ribaudo has opinions about this change.

nicolo-ribaudo (approved):
This seems good to me, I actually prefer the new approach of creating a fresh EC every time rather than just creating fresh environment records to attach to it.

ptomato:
Unfortunately this conflicted with #409 merged soon after your review @nicolo-ribaudo. I've rebased it and fixed the conflicts. Should be good to go, but I'd appreciate if you took another quick look at `new ShadowRealm()` where the conflict was.

caridy (approved):
LGTM

PR #339, opened 2021-11-02
Normative: Wrapped function should reuse target name and length

leobalter:
Closes #338

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -128,11 +128,14 @@ location: https://tc39.es/proposal-shadowrealm/
 		<h1>ShadowRealm Abstract Operations</h1>
 
 		<emu-clause id="sec-wrappedfunctioncreate" aoid="WrappedFunctionCreate">
-			<h1>WrappedFunctionCreate ( _callerRealm_, _Target_ )</h1>
+			<h1>
+				WrappedFunctionCreate (
+					_callerRealm_: a Realm Record,
+					_Target_: a function object,
+				)
+			</h1>
 			<p>The abstract operation WrappedFunctionCreate takes arguments _callerRealm_ and _Target_. It is used to specify the creation of new wrapped function exotic objects. It performs the following steps when called:</p>
 			<emu-alg>
-				1. Assert: _callerRealm_ is a Realm Record.
-				1. Assert: IsCallable(_Target_) is *true*.
 				1. Let _internalSlotsList_ be the internal slots listed in <emu-xref href="#table-internal-slots-of-wrapped-function-exotic-objects"></emu-xref>, plus [[Prototype]] and [[Extensible]].
 				1. Let _wrapped_ be ! MakeBasicObject(_internalSlotsList_).
 				1. Set _wrapped_.[[Prototype]] to _callerRealm_.[[Intrinsics]].[[%Function.prototype%]].
@@ -146,9 +149,15 @@ location: https://tc39.es/proposal-shadowrealm/
 		</emu-clause>
 
 		<emu-clause id="sec-copynameandlength" aoid="CopyNameAndLength">
-			<h1>CopyNameAndLength ( _F_, _Target_, _prefix_ [ , _argCount_ ] )</h1>
+			<h1>
+				CopyNameAndLength (
+					_F_: a function object,
+					_Target_: a function object,
+					_prefix_: a String,
+					optional _argCount_: a Number,
+				)
+			</h1>
 			<emu-alg>
-				1. Assert: Type(_prefix_) is String.
 				1. If _argCount_ is *undefined*, then set _argCount_ to 0.
 				1. Let _L_ be 0.
 				1. Let _targetHasLength_ be ? HasOwnProperty(_Target_, *"length"*).
--- a/spec.html
+++ b/spec.html
@@ -133,6 +133,20 @@ location: https://tc39.es/proposal-shadowrealm/
 			1. Set _wrapped_.[[Call]] as described in <emu-xref href="#sec-wrapped-function-exotic-objects-call-thisargument-argumentslist"></emu-xref>.
 			1. Set _wrapped_.[[WrappedTargetFunction]] to _Target_.
 			1. Set _wrapped_.[[Realm]] to _callerRealm_.
+			1. Let _L_ be 0.
+			1. Let targetHasLength be ? HasOwnProperty(_Target_, *"length"*).
+			1. If _targetHasLength_ is *true*, then
+				1. Let _targetLen_ be ? Get(_Target_, *"length"*).
+				1. If Type(_targetLen_) is Number, then
+					1. If _targetLen_ is *+&infin;*<sub>𝔽</sub>, set _L_ to +&infin;.
+					1. Else if _targetLen_ is *-&infin;*<sub>𝔽</sub>, set _L_ to 0.
+					1. Else,
+					1. Let _targetLenAsInt_ be ! ToIntegerOrInfinity(_targetLen_).
+						1. Assert: _targetLenAsInt_ is finite.
+						1. Set _L_ to max(_targetLenAsInt_, 0).
+			1. Perform ! SetFunctionLength(_wrapped_, _L_).
+			1. Let _name_ be ? Get(_Target_, *"name"*).
+			1. Perform ! SetFunctionName(_wrapped_, _name_, *"wrapped"*).
 			1. Return _wrapped_.
 		</emu-alg>
 	</emu-clause>
--- a/spec.html
+++ b/spec.html
@@ -120,43 +120,86 @@ location: https://tc39.es/proposal-shadowrealm/
 			In the case of an abrupt ~throw~ completion, the type of error to be created should match the type of the abrupt throw completion record. This could be revisited when merging into the main specification. Additionally, in the case of a ~break~ or ~continue~ completion, since those are not supported, a TypeError is expected.
 		</emu-note>
 	</emu-clause>
-
-	<emu-clause id="sec-wrappedfunctioncreate" aoid="WrappedFunctionCreate">
-		<h1>WrappedFunctionCreate ( _callerRealm_, _Target_ )</h1>
-		<p>The abstract operation WrappedFunctionCreate takes arguments _callerRealm_ and _Target_. It is used to specify the creation of new wrapped function exotic objects. It performs the following steps when called:</p>
-		<emu-alg>
-			1. Assert: _callerRealm_ is a Realm Record.
-			1. Assert: IsCallable(_Target_) is *true*.
-			1. Let _internalSlotsList_ be the internal slots listed in <emu-xref href="#table-internal-slots-of-wrapped-function-exotic-objects"></emu-xref>, plus [[Prototype]] and [[Extensible]].
-			1. Let _wrapped_ be ! MakeBasicObject(_internalSlotsList_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -122,18 +122,32 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
+			1. If Type(_targetLen_) is Number, then
+				1. If _targetLen_ is *+&infin;*<sub>𝔽</sub>, set _L_ to +&infin;.
+				1. Else if _targetLen_ is *-&infin;*<sub>𝔽</sub>, set _L_ to 0.
+				1. Else,
+				1. Let _targetLenAsInt_ be ! ToIntegerOrInfinity(_targetLen_).
+					1. Assert: _targetLenAsInt_ is finite.
+					1. Set _L_ to max(_targetLenAsInt_, 0).
+			1. Perform ! SetFunctionLength(_wrapped_, _L_).
+			1. Let _name_ be ? Get(_Target_, *"name"*).
+			1. Perform ! SetFunctionName(_wrapped_, _name_, *"wrapped"*).
```
legendecas, line 149: This makes the wrapping back and forth between several shadow realms observable from user code, like `"wrapped wrapped wrapped function"`. Is this behavior expected?

leobalter: I didn't think of that, and I'm not sure what's the best alternative other than just removing the prefix.

leobalter: It wouldn't be different to `Function#bind`

```js
function fn() {}
var bound = fn.bind().bind().bind();
bound.name; // 'bound bound bound fn'
```

legendecas: Right, it is the same with bind. I'm just confirming that this doesn't break any security assumptions.

ljharb, line 149: Note that `name` can be an object. Consider:
```js
function f() {}
Object.defineProperty(f, 'name', { value: f });
f.name === f
```

spec.html
```diff
@@ -122,18 +122,32 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
-			1. Return _obj_.
+			1. Let _wrapped_ be ! MakeBasicObject(_internalSlotsList_).
+			1. Set _wrapped_.[[Prototype]] to _callerRealm_.[[Intrinsics]].[[%Function.prototype%]].
+			1. Set _wrapped_.[[Call]] as described in <emu-xref href="#sec-wrapped-function-exotic-objects-call-thisargument-argumentslist"></emu-xref>.
+			1. Set _wrapped_.[[WrappedTargetFunction]] to _Target_.
+			1. Set _wrapped_.[[Realm]] to _callerRealm_.
+			1. Let _L_ be 0.
+			1. Let targetHasLength be ? HasOwnProperty(_Target_, *"length"*).
+			1. If _targetHasLength_ is *true*, then
+			1. Let _targetLen_ be ? Get(_Target_, *"length"*).
```
legendecas, line 139: Nit: indentation for steps nested with the condition `If _targetHasLength_ is *true*, then`?

spec.html
```diff
@@ -122,18 +122,32 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
+			1. If Type(_targetLen_) is Number, then
+				1. If _targetLen_ is *+&infin;*<sub>𝔽</sub>, set _L_ to +&infin;.
+				1. Else if _targetLen_ is *-&infin;*<sub>𝔽</sub>, set _L_ to 0.
+				1. Else,
+				1. Let _targetLenAsInt_ be ! ToIntegerOrInfinity(_targetLen_).
+					1. Assert: _targetLenAsInt_ is finite.
+					1. Set _L_ to max(_targetLenAsInt_, 0).
```
ljharb, line 146: These lines need to be indented.

spec.html
```diff
@@ -122,18 +122,32 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
+			1. Perform ! SetFunctionLength(_wrapped_, _L_).
```
ljharb, line 147: Since these exact steps are now repeated both here and in function bind, perhaps it should be factored into an AO?

spec.html
ljharb: Since anyone could at any time overwrite the name to be "wrapped wrapped wrapped anything", anyone with runtime assumptions about what a configurable name implies is already broken.

spec.html
leobalter: I'm fine doing that but I'd like to first define the final steps of this, as it seems like some tostring will be necessary inbetween.

ljharb: I don’t think converting to a string would make sense since bind doesn’t do that - i think you’d want to throw if it was a non-passable value.

Makes sense to wait for the refactor until you see what steps are shared.

spec.html
```diff
@@ -122,18 +122,33 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
+				1. If Type(_targetLen_) is Number, then
+					1. If _targetLen_ is *+&infin;*<sub>𝔽</sub>, set _L_ to +&infin;.
+					1. Else if _targetLen_ is *-&infin;*<sub>𝔽</sub>, set _L_ to 0.
+					1. Else,
+					1. Let _targetLenAsInt_ be ! ToIntegerOrInfinity(_targetLen_).
+						1. Assert: _targetLenAsInt_ is finite.
+						1. Set _L_ to max(_targetLenAsInt_, 0).
+			1. Perform ! SetFunctionLength(_wrapped_, _L_).
+			1. Let _targetName_ be ? Get(_Target_, *"name"*).
+			1. If Type(_targetName_) is not String, set _targetName_ to the empty String.
```
leobalter, line 149: @ljharb this is interesting! `Function#bind` normalizes to only copy string names, so I also applied this change here as you mentioned in the comments above.

There is a _catch_ where we need to avoid bringing error objects from the target function and that can happen on the HasOwnProperty and Get calls.

I believe that doing the common abstraction - as you suggested - would make it easier to handle this. I'll follow up with a new commit soon.

leobalter:
@ljharb PTAL

spec.html
```diff
@@ -120,29 +120,87 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
+				1. Set _wrapped_.[[WrappedTargetFunction]] to _Target_.
+				1. Set _wrapped_.[[Realm]] to _callerRealm_.
+				1. Let _result_ be CopyNameAndLength(_wrapped_, _Target_, *"wrapped"*).
+				1. If _result_ is an Abrupt Completion, throw a TypeError exception.
+				1. Return _wrapped_.
+			</emu-alg>
+		</emu-clause>
+
+		<emu-clause id="sec-copynameandlength" aoid="CopyNameAndLength">
+			<h1>CopyNameAndLength ( _F_, _Target_, _prefix_ [ , _argCount_ ] )</h1>
```
ljharb, line 149:
```suggestion
			<h1>
			  CopyNameAndLength (
			    _F_: a function object,
			    _Target_: a function object,
			    _prefix_: a String,
			    optional _argCount_: a Number,
			  )
			</h1>
```

spec.html
```diff
@@ -120,29 +120,87 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
+				1. Assert: _callerRealm_ is a Realm Record.
+				1. Assert: IsCallable(_Target_) is *true*.
+				1. Let _internalSlotsList_ be the internal slots listed in <emu-xref href="#table-internal-slots-of-wrapped-function-exotic-objects"></emu-xref>, plus [[Prototype]] and [[Extensible]].
+				1. Let _wrapped_ be ! MakeBasicObject(_internalSlotsList_).
+				1. Set _wrapped_.[[Prototype]] to _callerRealm_.[[Intrinsics]].[[%Function.prototype%]].
+				1. Set _wrapped_.[[Call]] as described in <emu-xref href="#sec-wrapped-function-exotic-objects-call-thisargument-argumentslist"></emu-xref>.
+				1. Set _wrapped_.[[WrappedTargetFunction]] to _Target_.
+				1. Set _wrapped_.[[Realm]] to _callerRealm_.
+				1. Let _result_ be CopyNameAndLength(_wrapped_, _Target_, *"wrapped"*).
+				1. If _result_ is an Abrupt Completion, throw a TypeError exception.
```
ljharb, line 143:
```suggestion
				1. Perform ? CopyNameAndLength(_wrapped_, _Target_, *"wrapped"*).
```
currently this is catching the almost-certainly-TypeError and throwing a TypeError, which would be masking the actual cause

spec.html
```diff
@@ -120,29 +120,87 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
...
-	</emu-clause>
 </emu-clause>
 
 <emu-clause id="sec-shadowrealm-objects">
 	<h1>ShadowRealm Objects</h1>
 	<emu-clause id="sec-shadowrealm-abstracts">
 		<h1>ShadowRealm Abstract Operations</h1>
 
+		<emu-clause id="sec-wrappedfunctioncreate" aoid="WrappedFunctionCreate">
+			<h1>WrappedFunctionCreate ( _callerRealm_, _Target_ )</h1>
```
ljharb, line 131:
```suggestion
			<h1>
			  WrappedFunctionCreate (
			    _callerRealm_: a Realm Record,
			    _Target_: an ECMAScript language value,
			  )
			</h1>
```
and then the assertion on line 134 can be removed.

spec.html
leobalter: is this a new notation?

spec.html
leobalter: the issue here is not with TypeError, is not leaking cross realms error objects.

spec.html
ljharb: yes, the entire spec uses it for AOs since https://github.com/tc39/ecma262/pull/545 was merged in July.

spec.html
ljharb: ah, i see. shouldn't the message, or the cause, or the type (if it's a builtin) be preserved if possible?

leobalter: there is no definition for the message, same as in `PerformShadowRealmEval`. 

There is no logic yet for the cause, and I'd prefer to do that in a separate PR and tackle all these points throwing TypeError.

leobalter:
@ljharb we have the formatted header now. Is there anything else for this PR?

spec.html
caridy: btw, the spec was written in a way that implementers can optimize the wrapping to wrap the original target rather than the wrapped one to avoid the different jumps, and we should preserve that. And I believe we should be careful to preserve that.

mhofman: Prefixing should prevent any call optimization, right? The name is defined at wrapping time, where the target resolution is at call time.

Jack-Works:
What's the prior art of this change? If it's performance consideration, I think it should be host defined, host either copy the name and length or do nothings

mhofman:
I'd be opposed to host defined behavior. We want more predictable behavior, not less

ljharb (approved):

jdalton:
I like the being similar to `bind` approach too.

caridy (approved):
LGTM

leobalter:
This PR should observe any outcome from https://github.com/tc39/proposal-shadowrealm/issues/338#issuecomment-964606972

leobalter:
This achieved consensus during the [TC39 plenary in Dec 2021](https://github.com/tc39/notes/blob/master/meetings/2021-12/dec-14.md#shadowrealms-updates-and-potential-normative-changes).

PR #346, opened 2022-01-19
Editorial: update AO parameter declaration notations

legendecas:
Removes the redundant steps of type assertions.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -96,9 +96,17 @@ location: https://tc39.es/proposal-shadowrealm/
 	</table>
 	</emu-table>
 
-	<emu-clause id="sec-wrapped-function-exotic-objects-call-thisargument-argumentslist">
-		<h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
-		<p>The [[Call]] internal method of a wrapped function exotic object _F_ takes arguments _thisArgument_ (an ECMAScript language value) and _argumentsList_ (a List of ECMAScript language values). It performs the following steps when called:</p>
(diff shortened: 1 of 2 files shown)


spec.html
```diff
@@ -121,11 +129,18 @@ <h1>[[Call]] ( _thisArgument_, _argumentsList_ )</h1>
 		</emu-note>
 	</emu-clause>
 
-	<emu-clause id="sec-wrappedfunctioncreate" aoid="WrappedFunctionCreate">
-		<h1>WrappedFunctionCreate ( _callerRealm_, _targetFunction_ )</h1>
-		<p>The abstract operation WrappedFunctionCreate takes arguments _callerRealm_ and _targetFunction_. It is used to specify the creation of new wrapped function exotic objects. It performs the following steps when called:</p>
+	<emu-clause id="sec-wrappedfunctioncreate" type="abstract operation">
```
leobalter, line 132: is the aoid not needed here anymore?

leobalter (commented):
LGTM with a quick question. Thanks for doing this work!

spec.html
legendecas: ecmarkup suggested that if the header is structured then the aoid is no longer needed.

legendecas: Maybe I'm mistaken, I didn't find ecmarkup warning about this again :\. Reverted.

leobalter: It's all fine! I honestly don't know the correct answer and appreciate your attention to details here.

PR #349, opened 2022-02-24
Editorial: GetFunctionRealm may throw on revoked proxies

legendecas:
Hoist the note so that the thrown TypeError on revoked proxies are thrown in the caller realm.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -110,9 +110,9 @@ location: https://tc39.es/proposal-shadowrealm/
 		<emu-alg>
 			1. Let _target_ be _F_.[[WrappedTargetFunction]].
 			1. Assert: IsCallable(_target_) is *true*.
-			1. Let _targetRealm_ be ? GetFunctionRealm(_target_).
-			1. Let _callerRealm_ be ? GetFunctionRealm(_F_).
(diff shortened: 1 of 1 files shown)


legendecas:
friendly ping @rwaldron @caridy :)

rwaldron (approved):

leobalter:
As Rick mentioned in another channel, this seems too fine so we worry if we're not missing any effects. I agree with this change, I just want to wait a bit for @caridy to take a look as well. He's currently on a time off so I can't say when this is happening, but expect soon.

caridy (commented):
@legendecas LGTM

I believe the observable difference here is related to revoked proxies coming from a different window (e.g.: a proxy created by a same-domain iframe that was revoked, and pass into a shadow realm associated to the main window. The difference now is that the error is from the main window (caller realm) rather than the iframe's realm. It makes the code more predictable IMO.

caridy (approved):
👍

leobalter:
@ljharb will eventually remind me this represents a normative change so I'll report this at the TC39 meeting as well, but I believe it's not gonna be an issue.

PR #352, opened 2022-03-11
Normative: Type check on importValue 2nd value

Jack-Works:
close #350


Jack-Works:
I think we can add this to the next meeting too.

leobalter:
This achieved consensus at TC39 plenary in March 2022.

leobalter (approved):

PR #373, opened 2022-09-01
Editorial: fix #371 - better handling of completion values

caridy:


caridy:
cc @linusg

spec.html
```diff
@@ -111,10 +111,10 @@ <h1>
 			1. Let _callerContext_ be the running execution context.
 			1. Let _calleeContext_ be PrepareForWrappedFunctionCall(_F_).
 			1. Assert: _calleeContext_ is now the running execution context.
-			1. Let result be OrdinaryWrappedFunctionCall(_F_, _thisArgument_, _argumentsList_).
+			1. Let _result_ be Completion(OrdinaryWrappedFunctionCall(_F_, _thisArgument_, _argumentsList_)).
```
caridy, line 114: Since `OrdinaryWrappedFunctionCall` can throw... at multiple places... better to be sure here.

spec.html
```diff
@@ -111,10 +111,10 @@ <h1>
...
 			1. Remove _calleeContext_ from the execution context stack and restore _callerContext_ as the running execution context.
-			1. If result.[[Type]] is return, return result.[[Value]].
-			1. ReturnIfAbrupt(result).
+			1. ReturnIfAbrupt(_result_).
+			1. If _result_.[[Type]] is |return|, return _result_.[[Value]].
```
linusg, line 117: Unfortunately I don't see how this fixes the issue:

- As stated in the original issue, `OrdinaryWrappedFunctionCall` can't produce a completion of `[[Type]]` _return_ AFAICT, wrapping it in a `Completion()` call doesn't change that fact, only makes it more explicit - https://tc39.es/ecma262/#sec-completion-ao:
 > The abstract operation Completion takes argument completionRecord (a [Completion Record](https://tc39.es/ecma262/#sec-completion-record-specification-type)) and returns a [Completion Record](https://tc39.es/ecma262/#sec-completion-record-specification-type). It is used to emphasize that a [Completion Record](https://tc39.es/ecma262/#sec-completion-record-specification-type) is being returned.
- `ReturnIfAbrupt` returns if the completion type is anything other than _normal_, which includes _return_, so the next line is unreachable

linusg:
In case it is helpful, what I did in my implementation to work around the spec bug is simply `Return result.`:

https://github.com/SerenityOS/serenity/blob/dfb7588d300148ec72e1283947c4b37d71ffc460/Userland/Libraries/LibJS/Runtime/WrappedFunction.cpp#L75-L79

I was assuming that:
- Step 6 was supposed to handle normal completions, otherwise the [[Value]] from `OrdinaryWrappedFunctionCall` would be completely dropped in favor of returning `undefined`
- Step 7 would make step 8 redundant as only a normal or throw completion could be returned, and throw is handled by `ReturnIfAbrupt`

This implementation of ShadowRealm passes all 64 tests currently present in test262: https://libjs.dev/test262/per-file/?path=test/built-ins/ShadowRealm

So in case the desired behaviour is something else, that is currently not convered by tests (or unobservable).

leobalter:
@rwaldron PTAL at the referred test262.

PR #374, opened 2022-09-02
Editorial: Remove unused evalRealm parameter from ShadowRealmImportValue

linusg:


spec.html
```diff
@@ -316,7 +316,6 @@ <h1>
 					_specifierString_: a String,
 					_exportNameString_: a String,
 					_callerRealm_: a Realm Record,
-					_evalRealm_: a Realm Record,
 					_evalContext_: an execution context,
```
legendecas, line 319: What about removing the parameter `_evalContext_` and the first assertion of `ShadowRealmImportValue` together instead? This way it can be more consistent with `PerformShadowRealmEval(_sourceText_, _callerRealm_, _evalRealm_)`.

linusg: But `evalContext` is pushed before calling `HostImportModuleDynamically`, I don't think we can remove it

legendecas: Yeah, you are right about this.

legendecas (approved):
LGTM

PR #381, opened 2022-11-18
Adding explainer for error.message and censoring error.stack

caridy:
This PR introduces an extended explainer for `error.message` and `error.stack`.


littledan (commented):
It would be nice if this document started with the basics:
- Errors thrown across the Realm boundary in either direction are replaced by a fresh TypeError
- This TypeError copies the `.message` property iff that is a string
- We encourage engines to *enable* their engine-specific stack introspection mechanisms to work on such TypeErrors, subject to the following censorship constraints.

errors.md
```diff
@@ -0,0 +1,134 @@
...
+
+Now, let's assume that we have 3 programs:
+
+ - Main Program (running on top level)
+   - Program A (running inside a ShadowRealm created by the Main Program)
+     - Program B (running inside a ShadowRealm created by Program A)
+
+The same mechanism must be applied if the error was generated by a child ShadowRealm B (Program B), which was created by Program A. Let's explore the different scenarios:
+
+1. A try/catch in Program B catches the error, it will qualify to be censored, only revealing stack frames associated to the ShadowRealm where Program A runs, and none from Program B or the Main Program.
```
bathos, line 127: This is a typo, right?

```suggestion
1. A try/catch in Program B catches the error, it will qualify to be censored, only revealing stack frames associated to the ShadowRealm where Program B runs, and none from Program A or the Main Program.
```

bathos:
If lowercase names represent incubator realm functions and uppercase those of one shadow realm in `a(B(c(D(e()))))` and e introspects `new Error().stack` (nothing is thrown), does it see a-c-e or a-B-c-D-e now? It seems like the text here does not say for certain, but it might imply the latter? I may be misunderstanding, though.

(FWIW if `e` can now observe `B` and `D`’s interleaved presence at will like that, ShadowRealm no longer satisfies my use cases.)

mhofman:
The intent was to allow a-B-c-D-e. The assumption is that the incubator realm is in a position to wrap ShadowRealm it constructs anyway, so it'd be able to stitch stacks.

I am curious about your use case. Would you be able to share why an incubator realm shouldn't be able to automatically introspect these stack frames? We were trying to strike a balance, and heard feedback that requiring things like telemetry libraries in the incubator realm to recursively wrap the ShadowRealm constructors to stitch stacks was too onerous.

bathos:
> Would you be able to share why an incubator realm shouldn't be able to automatically introspect these stack frames?

Implementation and virtualization of “native” Web IDL constructs. The incubator realm (or other realms that share first-class object values) includes the bootstrapping code / interfaces over for the implementation layer but also consumer code: the ShadowRealm is providing encapsulation for the implementation code rather than for evaluation of arbitrary or otherwise “untrusted” (or whatever term is accepted — you know what I mean I’m sure) code.

> heard feedback that requiring things like telemetry libraries in the incubator realm to recursively wrap the ShadowRealm constructors to stitch stacks was too onerous.

It is not the exposure that is the problem but the conditions of exposure: it would need to occur only when a genuine throw completion crossed the boundary. If the error were _thrown_ from the shadow, it would be no problem if the stack included all frames because it would represent a real error in the virtualized implementation layer.

errors.md
caridy: Yes, a typo. good catch @bathos

caridy:
> If lowercase names represent incubator realm functions and uppercase those of one shadow realm in a call stack like `a → B → c → D → e` and e includes code introspecting `new Error().stack` (note nothing is thrown), does it “see” a-c-e or a-B-c-D-e now? It seems like the text here does not say for certain, but it might imply the latter? I may be misunderstanding, though.

@bathos It is the latter `a-B-c-D-e` but only if the incubator realm is not a ShadowRealm (say it is a window). If, on the other hand, the incubator realm is a ShadowRealm, e.g.: Window->ShadowRealm_X->ShadowRealm_Y`, then only code in Window can see the full stack, and therefore `e` was evaluated in ShadowRealm_X, and the stack property will contain `a-c-e` because the other frames do not belong to the realm associated to the ShadowRealm instance were `e` is evaluated. It is about individual stack frames, and they associated Realm only when observed from ShadowRealms.

littledan (approved):
The new introduction is very well-written, though it doesn't match the new spec text. I guess we could just delete the normative text on `message` and encourage engines to do something like this, more or less. The censorship goal will require consensus.

caridy:
> The new introduction is very well-written, though it doesn't match the new spec text. I guess we could just delete the normative text on `message` and encourage engines to do something like this, more or less. The censorship goal will require consensus.

Agreed! I have updated the spec to remove those steps, and I have added a lot more details here about the suggested process.

errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+New `Error.message` after crossing a boundary:
+
+```
+Uncaught TypeError: wrapped function threw, error was TypeError: null has no properties
+```
+
+This error allows developers to clearly understand that the error was thrown from another Realm. If the error crosses multiple nested ShadowRealms, the second time the error is created when crossing another boundary, the message should still be formed from scratch rather than providing nesting of the message. The same applies to re-entrancing as well since the error re-entring the ShadowRealm where it was originated, will come as a brand new TypeError object, with no connection to the original Error object.
+
+Accessing the `name` and `message` of the original Error object must not be observed by the user-land code. The following logic may be used:
```
caridy, line 27: @littledan @mhofman my realization was that FF already implemented this logic (I didn't know about it). But it is sound IMO. Please, take a look.

errors.md
```diff
@@ -0,0 +1,156 @@
...
+Original `Error.message`:
+
+```
+TypeError: null has no properties
+```
+
+New `Error.message` after crossing a boundary:
+
+```
+Uncaught TypeError: wrapped function threw, error was TypeError: null has no properties
```
mhofman, line 22:
```suggestion
TypeError: wrapped function threw, error was TypeError: null has no properties
```

errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Errors Crossing Callable Boundary
+
+Errors thrown across the ShadowRealm's callable boundary in either direction are replaced by a fresh `TypeError` as described by the spec. Additionally, the new `TypeError` instance can be augmented with a `message` and `stack` properties to help developers.
+
+### New Error.message
+
+The new `TypeError` object created to propagate an error across the callable boundary can be augmented with a `message` property. The value of such property can be a combination of `name` and `message` from the original error. E.g.:
+
+Original `Error.message`:
```
mhofman, line 13:
```suggestion
Original `error.message`:
```

Also an original `new TypeError('foo').message === 'foo'`, but `String(new TypeError('foo')) === 'TypeError: foo'`

errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+```
+TypeError: null has no properties
+```
+
+New `Error.message` after crossing a boundary:
```
mhofman, line 19:
```suggestion
New `error.message` after crossing a boundary:
```

errors.md
```diff
@@ -0,0 +1,156 @@
...
+Uncaught TypeError: wrapped function threw, error was TypeError: null has no properties
+```
+
+This error allows developers to clearly understand that the error was thrown from another Realm. If the error crosses multiple nested ShadowRealms, the second time the error is created when crossing another boundary, the message should still be formed from scratch rather than providing nesting of the message. The same applies to re-entrancing as well since the error re-entring the ShadowRealm where it was originated, will come as a brand new TypeError object, with no connection to the original Error object.
+
+Accessing the `name` and `message` of the original Error object must not be observed by the user-land code. The following logic may be used:
+
+If _originalError_ has an `[[ErrorData]]` internal slot, then use the data values of `name` and `message` of the data properties. In case of accesor properties, use the cached values stored during the creation of the original error by the host.
+
+_Note: It is possiable to only cache the `name` and `message` if the error could cross a boundary (based on the stack frames)._
```
mhofman, line 31:
```suggestion
_Note: It is possible to only cache the `name` and `message` if the error could cross a boundary (based on the stack frames)._
```

errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+It produces a TypeError with the following message:
+
+```
+wrapped function threw, error was uncaught exception: Object
+```
+
+_Note: FF implementation follows the logic described above._
+
+### New Error.Stacks
```
mhofman, line 49:
```suggestion
### error.stack
```

errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+A good error message is sometimes not enough. We encourage engines to enable their engine-specific stack introspection mechanisms for errors in ShadowRealms. In this case, the Error's stack is subject to censorship constraints.
+
+Some of the use-cases for ShadowRealm, e.g., the virtualization, require a mechanism to control the error's stack, so those accessing it should not observe that the program is running inside a ShadowRealm. Early versions of the spec were simply nulling out the stack when an error crosses the callable boundary, but that was not sufficient, in fact, it was faulty for two main reasons:
```
mhofman, line 53: > the virtualization

of ? another host?

errors.md
```diff
@@ -0,0 +1,156 @@
...
+```js
+try { null.foo } catch (e) {
+    console.log(e.stack);
+    throw e;
+}
+```
+
+In the example above, `e` never crossed a callable boundary, but regardless, it should not contain a stack frame associated to another realm. Assuming that this error is observed (accessed) in two different places, a try/catch from above when running inside the ShadowRealm instances, and another try/catch in the incubator realm, which happens to be the main page (top level window), we should see two different error's stack for the two different errors:
+
+1. ShadowRoot's `Error.stack`:
```
mhofman, line 73:
```suggestion
1. ShadowRealm's `error.stack`:
```

errors.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+_Note: Only the 3 stack frames must be exposed on this error because they correspond to code evaluated inside the ShadowRealm associated to the error instance._
+
+2. Main page's `Error.stack`:
+```
+foo@http://127.0.0.1:8081/y.js:2:5
+bar@http://127.0.0.1:8081/y.js:6:12
+captureStack@http://127.0.0.1:8081/x.js:9:9
+runInSandbox@http://127.0.0.1:8081/demo.html:6:17
```
mhofman, line 87: nit but `runInSanbox` may be distracting here, and a `runVirtualized` or something like that may be more appropriate.

errors.md
```diff
@@ -0,0 +1,156 @@
...
+@http://127.0.0.1:8081/demo.html:8:1
+```
+
+_Note: All 5 stack frames must be exposed on this error regardless of their location and associated Realms, this is analogous to what happen with same domain iframes today._
```
mhofman, line 91:
```suggestion
_Note: All 5 stack frames should be exposed on this error regardless of their location and associated Realms, this is analogous to what happen with same domain iframes today._
```

explainer.md
```diff
@@ -388,6 +388,10 @@ function extractIntrinsicsFromGlobal(customGlobalThis) {
 }
 ```
 
+## <a name='Errors'></a>Errors
+
+Errors originated from a ShadowRoot are subject to stack censoring. Similarely, those errors must be copied when crossing the callable boundary, while doing so, the host must produce a TypeError, and may provide a message and stack properties without violating the stack censoring principle. For more details about errors, refer to the [errors explainer](./errors.md).
```
mhofman, line 393:
```suggestion
Errors originated from a ShadowRealm are subject to stack censoring. Similarly, those errors must be copied when crossing the callable boundary, while doing so, the host must produce a TypeError, and may provide a message and stack properties without violating the stack censoring principle. For more details about errors, refer to the [errors explainer](./errors.md).
```

mhofman (approved):
LGTM. Some nits about formatting and capitalization.

PR #391, opened 2023-12-15
Editorial: Use consistent Well-Known capitalization as ECMAScript Spec

arai-a:
ECMAScript spec used "Well-Known" as capitalization in the header and caption.

https://tc39.es/ecma262/#sec-well-known-intrinsic-objects


ptomato (approved):
Thanks!

PR #392, opened 2023-12-15
Let the host create the global object

Ms2ger:
Ref https://github.com/whatwg/html/issues/7591

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -472,7 +472,7 @@ location: https://tc39.es/proposal-shadowrealm/
 			<emu-alg>
 				1. If NewTarget is *undefined*, throw a *TypeError* exception.
 				1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%ShadowRealm.prototype%"*, « [[ShadowRealm]], [[ExecutionContext]] »).
-				1. Perform ? InitializeHostDefinedRealm(). The customizations for creating the global object are to return ? InitializeShadowRealm().
+				1. Perform ? InitializeHostDefinedRealm(). The customizations for creating the global object are to return ? InitializeShadowRealm(_O_).
 				1. Let _context_ be the running JavaScript execution context.
 				1. Set _O_.[[ExecutionContext]] to _context_.
 				1. Let _realmRec_ be the Realm of _context_.
@@ -481,13 +481,16 @@ location: https://tc39.es/proposal-shadowrealm/
 			</emu-alg>
 
 			<emu-clause id="sec-initializeshadowrealm" type="abstract operation">
-				<h1>InitializeShadowRealm ( ): either a normal completion containing an Object or a throw completion</h1>
+				<h1>
+					InitializeShadowRealm (
+						_O_: a ShadowRealm object
+					): either a normal completion containing an Object or a throw completion</h1>
 				<dl class="header">
 				</dl>
 				<emu-alg>
 					1. Let _context_ be the running JavaScript execution context.
 					1. Let _realm_ be the Realm of _context_.
-					1. Return ? HostInitializeShadowRealm(_realm_, _context_).
+					1. Return ? HostInitializeShadowRealm(_realm_, _context_, _O_).
 				</emu-alg>
 			</emu-clause>
 		</emu-clause>
@@ -596,6 +599,7 @@ location: https://tc39.es/proposal-shadowrealm/
 				HostInitializeShadowRealm (
 					_realm_: a Realm Record,
 					_context_: an execution context,
+					_O_: a ShadowRealm object,
 				): either a normal completion containing an Object or a throw completion
 			</h1>
 			<dl class="header">
--- a/spec.html
+++ b/spec.html
@@ -479,7 +479,7 @@ location: https://tc39.es/proposal-shadowrealm/
 				1. Set the Realm of _context_ to _realmRec_.
 				1. Set the ScriptOrModule of _context_ to *null*.
 				1. Set _O_.[[ExecutionContext]] to _context_.
-				1. Perform ? HostInitializeShadowRealm(_O_.[[ShadowRealm]]).
+				1. Perform ? HostInitializeShadowRealm(_O_.[[ShadowRealm]], _O_.[[ExecutionContext]]).
 				1. Return _O_.
 			</emu-alg>
 		</emu-clause>
@@ -587,6 +587,7 @@ location: https://tc39.es/proposal-shadowrealm/
 			<h1>
 				HostInitializeShadowRealm (
 					_realm_: a Realm Record,
+					_context_: an execution context,
 				): either a normal completion containing ~unused~ or a throw completion
 			</h1>
 			<dl class="header">
--- a/spec.html
+++ b/spec.html
@@ -479,8 +479,6 @@ location: https://tc39.es/proposal-shadowrealm/
 				1. Set the Realm of _context_ to _realmRec_.
 				1. Set the ScriptOrModule of _context_ to *null*.
 				1. Set _O_.[[ExecutionContext]] to _context_.
-				1. Perform SetRealmGlobalObject(_realmRec_, *undefined*, *undefined*).
-				1. Perform ? SetDefaultGlobalBindings(_O_.[[ShadowRealm]]).
 				1. Perform ? HostInitializeShadowRealm(_O_.[[ShadowRealm]]).
 				1. Return _O_.
 			</emu-alg>
@@ -600,6 +598,12 @@ location: https://tc39.es/proposal-shadowrealm/
 					to the ShadowRealm, e.g., for module loading.
 				</dd>
 			</dl>
+			<p>The default implementation of HostInitializeShadowRealm performs the following steps when called:</p>
+			<emu-alg>
+				1. Perform SetRealmGlobalObject(_realm_, *undefined*, *undefined*).
+				1. Perform ? SetDefaultGlobalBindings(_realm_).
+				1. Return ~unused~.
+			</emu-alg>
 			<p>
 				The host may use this hook to add properties to the ShadowRealm's global
 				object. Those properties must be configurable.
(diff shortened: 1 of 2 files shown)


spec.html
```diff
@@ -612,11 +616,9 @@ <h1>
 					including `HTMLElement`, `console`, `localStorage`, `fetch`, etc.
 				</p>
 			</emu-note>
-			<emu-note type=editor>
+			<emu-note>
 				<p>
-					The ShadowRealm constructor (<emu-xref href="#sec-shadowrealm"></emu-xref>)
-					creates a new global object as an ordinary object. This
-					means all properties from the global object are deletable.
+					The host is expected to create global object whose properties are all deletable.
```
linusg, line 621:
```suggestion
					The host is expected to create a global object whose properties are all deletable.
```

Although I think this note is redundant, the paragraph above already states:

> The host may use this hook to add properties to the ShadowRealm's global object. Those properties must be configurable.

...and configurable means deletable.

mhofman:
This change looks like a regression. It seems that the host gains the ability to create a global object with exotic behavior. In particular the main requirement is that all properties must be effectively deletable. I also believe the intent was for the prototype of the global object to be effectively mutable, and that the global object could be successfully frozen. The reason I mention "effectively" is that an exotic object can claim that properties are configurable, yet refuse to actually mutate or delete them.

Ms2ger:
I'm happy to add restrictions on the host, but would prefer if someone with more background suggested prose for it.

mhofman:
#394 raises an interesting point about discovery of the host added properties. If the host creates the object, it should implement all added properties as own properties.

I think basically the constraint should be that all behavior implemented by the host created global object be observably regular / non-exotic. In this case the object invariants are not sufficient as they provide too much leeway. At which point I'm still curious why we need this change since a host created object with regular object behavior, and a spec created regular object are observably indistinguishable, and can be considered an internal implementation concern.

spec.html
leobalter: @linusg with historical facts from Ecmascript and Web, it may seem obvious, but not necessarily. Hence @mhofman - along with the proposal champions - emphasizing on the Global being an ordinary object.

caridy (approved):
LGTM

ByteEater-pl:
@mhofman, could @ljharb's ["Get Intrinsic" Proposal](https://github.com/tc39/proposal-get-intrinsic) help reconcile these concerns?

We could use it to enumerate the global object's properties in a ShadowRealm and delete all or some of them (because they're required to be configurable, though that requirement would have to be strengthened so that the deletability isn't faked), e.g. just those added by the host.

And hosts with `EventTarget` wishing to have ShadowRealms' global objects inherit from it would be free to do so.

Jack-Works:
I oppose to host creating the global object. `HostInitializeShadowRealm` is enough.

Without userland ShadowRealm API to do the same thing, an exotic global object is unacceptable.

Can you give the reason why the host should take over the creation?

mhofman:
> could @ljharb's ["Get Intrinsic" Proposal](https://github.com/tc39/proposal-get-intrinsic) help reconcile these concerns?

I don't see how that proposal would help. The concern is not with the ability to enumerate globals, but with requiring that the global object has non-exotic behavior.



> And hosts with `EventTarget` wishing to have ShadowRealms' global objects inherit from it would be free to do so.

Why is the host creating the global object required for this? The prototype of a fresh regular object is mutable. Nothing in the original steps says the global object has to inherit from `Object.prototype`. Also as I mentioned, whether in the actual implementation the host does in fact create the global object is unobservable, as long as the global object can be observed to have no exotic behavior.

Jack-Works:
The host can set the global object's prototype to EventTarget.prototype in `HostInitializeShadowRealm`. The host can also attach the required internal slot of EventTarget to make it functional, this is virtualizable since a non-Web implementation can also do that by using their own fake "EventTarget" that accepts the global object as a valid `this`.

ByteEater-pl:
@mhofman, you wrote:
> If the host creates the object, it should implement all added properties as own properties.

If instead it creates a regular object as global and operates on it in `HostInitializeShadowRealm` ([whereby](https://tc39.es/proposal-shadowrealm/#sec-hostinitializeshadowrealm) setting its prototype, not just adding properties, should therefore be mentioned as an example of what's allowed and likely useful), what stops it from setting an exotic object as a prototype? If it does, and "use this hook to add properties to the ShadowRealm's [global object](https://tc39.es/ecma262/#sec-global-object)" is interpreted (correctly, as we both seem to assume) as including inherited properties (to which the configurability requirement applies), we're back in square one, since the prototype (or any object up in the prototype chain) can:

- fake delibility,
- effectively deny enumerability.

What I mean by the (actually understated; worse things can happen) latter is, assuming that in the absence of [Get Intrinsic](https://github.com/tc39/proposal-get-intrinsic) (which I therefore mentioned) you'd traverse the prototype chain with `getPrototypeOf` (on `Object` or `Reflect`) and enumerate everything available, some object on the prototype chain may have a custom `[[GetPrototypeOf]]`, which opens possibilities (not prevented by invariants of the essential internal methods) such as returning a different prototype to the enumerating code than in normal operation, sporting or lazily creating an infinite prototype chain, adding properties (even non-configurable ones) to the global object itself (or any ancestor in the prototype chain), throwing, or simply not terminating.

ByteEater-pl:
Or maybe it's fine to change the prototype of the global object first (it's ordinary, so possible), thereby loosing its `EventTarget` quality, and then turn to deleting what you don't want on the object itself.

But even then, I believe the requirements of `HostInitializeShadowRealm` should be strengthened. If it's currently understood that it can set the global object's prototype, it probably can also delete, replace or reconfigure (even make non-configurable) the properties installed by `SetDefaultGlobalBindings`. In this and other ShadowRealms. And mangle your module graph (outside the ShadowRealm). And make demons fly out of your nose. Yea, I know other host operations aren't usually so carefully limited, but this one is special, security hinges on it.

mhofman:
> what stops it from setting an exotic object as a prototype?

That's fine, as long as the global object remains extensible, the prototype object can be substituted, which is sufficient.

> including inherited properties (to which the configurability requirement applies)

The configurability requirement does not apply to the prototype chain, only to the global object.

> you'd traverse the prototype chain with `getPrototypeOf` (on `Object` or `Reflect`) and enumerate everything available

Nope, you'd likely replace the prototype of the global object with a wrapper or other object that behaves as expected.

> Or maybe it's fine to change the prototype of the global object first (it's ordinary, so possible), thereby loosing its `EventTarget` quality, and then turn to deleting what you don't want on the object itself.

Right, but the user doesn't have to lose the `EventTarget` methods, that prototype can be inserted in the chain, or a wrapper for it, etc.


> But even then, I believe the requirements of `HostInitializeShadowRealm` should be strengthened.

> even make non-configurable) the properties installed by `SetDefaultGlobalBindings`

I agree, `HostInitializeShadowRealm` should not be allowed to add or make any properties of the global object non-configurable, nor should it be allowed to make the global object non-extensible.

> In this and other ShadowRealms

Yes, I wanted to have some text somewhere that prevented the host from making properties appear or otherwise change properties on the global object after `HostInitializeShadowRealm`, but I can't track down where I asked for this.

> And mangle your module graph (outside the ShadowRealm). And make demons fly out of your nose. Yea, I know other host operations aren't usually so carefully limited, but this one is special, security hinges on it.

I honestly don't care about what the host does outside the shadow realm after its creation, as long as it respects the callable boundary invariants (which are now spelled out precisely) and the above requirements to not further change the global object after initialization.

Jack-Works:
> we're back in square one, since the prototype (or any object up in the prototype chain) can:

>    fake delibility,
>    effectively deny enumerability.

that's ok, because you can do it in the user land with Proxy. but you cannot make the global object _itself_ becomes a Proxy.

> If it's currently understood that it can set the global object's prototype, it probably can also delete, replace or reconfigure (even make non-configurable) the properties installed by SetDefaultGlobalBindings.

I agree. It is (if not, it must is) a formal requirement to implementations that all properties are configurable, even if they installed it in `HostInitializeShadowRealm`.

spec.html
```diff
@@ -472,18 +472,27 @@ <h1>ShadowRealm ( )</h1>
...
-				1. Set the Function of _context_ to *null*.
-				1. Set the Realm of _context_ to _realmRec_.
-				1. Set the ScriptOrModule of _context_ to *null*.
+				1. Perform ? InitializeHostDefinedRealm(). The customizations for creating the global object are to return ? InitializeShadowRealm(_O_).
+				1. Let _context_ be the running JavaScript execution context.
 				1. Set _O_.[[ExecutionContext]] to _context_.
-				1. Perform SetRealmGlobalObject(_realmRec_, *undefined*, *undefined*).
-				1. Perform ? SetDefaultGlobalBindings(_O_.[[ShadowRealm]]).
-				1. Perform ? HostInitializeShadowRealm(_O_.[[ShadowRealm]]).
+				1. Let _realmRec_ be the Realm of _context_.
```
caridy, line 478: @Ms2ger I'm confused about this line, and the title of this PR. Who is creating the realm now?

linusg: `CreateRealm` was folded into `InitializeHostDefinedRealm` in https://github.com/tc39/ecma262/pull/3139, it creates a realm, customizable global object, and execution context which is pushed to the stack. This may be changed again slightly in https://github.com/tc39/ecma262/pull/3139.

mhofman: I'm also a little confused, and still don't understand the purpose of this PR. For ShadowRealm we need a global object with observably non-exotic behavior. Why can't we just let the 262 pretend it created a regular object to use as global, and have implementations actually implement it differently if they want to have their global object with exotic behavior that is non observable?

mhofman: In https://github.com/tc39/ecma262/pull/3139 I see the following: 
> If the host requires use of an exotic object to serve as _realm_'s global object, then

Can we in any way constraint here that the host is not allowed to use an exotic object for the global object of ShadowRealm? And thus pretend that observably we took the path of the Else and created a regular object for the global?

erights (changes requested):
I agree with @mhofman 's comments in this PR's conversations.

Ms2ger:
I'll come back to this after  #409

caridy:
@Ms2ger now that #409 is closed, can we get back to this and figure out what to do?

Ms2ger:
I'm looking into this; at this point it seems like this PR might not be necessary.

mhofman:
> I'm looking into this; at this point it seems like this PR might not be necessary.

That's because other PRs completely absorbed the original change from this PR. However the concerns I raise are valid more than ever. I will open an issue that must be addressed before proceeding to stage 3.

mhofman:
Filed https://github.com/tc39/proposal-shadowrealm/issues/411

PR #409, opened 2024-10-24
Rewrite the ShadowRealm constructor with InitializeHostDefinedRealm

Ms2ger:
The current prose no longer works after https://github.com/tc39/ecma262/pull/3139.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -472,16 +472,12 @@ location: https://tc39.es/proposal-shadowrealm/
 			<emu-alg>
 				1. If NewTarget is *undefined*, throw a *TypeError* exception.
 				1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%ShadowRealm.prototype%"*, « [[ShadowRealm]], [[ExecutionContext]] »).
-				1. Let _realmRec_ be CreateRealm().
-				1. Set _O_.[[ShadowRealm]] to _realmRec_.
-				1. Let _context_ be a new execution context.
-				1. Set the Function of _context_ to *null*.
-				1. Set the Realm of _context_ to _realmRec_.
-				1. Set the ScriptOrModule of _context_ to *null*.
+				1. Perform ? InitializeHostDefinedRealm().
+				1. Let _context_ be the running execution context.
 				1. Set _O_.[[ExecutionContext]] to _context_.
-				1. Perform SetRealmGlobalObject(_realmRec_, *undefined*, *undefined*).
-				1. Perform ? SetDefaultGlobalBindings(_O_.[[ShadowRealm]]).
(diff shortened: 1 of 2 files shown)


spec.html
```diff
@@ -472,16 +472,12 @@ <h1>ShadowRealm ( )</h1>
...
-				1. Set the ScriptOrModule of _context_ to *null*.
+				1. Perform ? InitializeHostDefinedRealm().
+				1. Let _context_ be the running execution context.
 				1. Set _O_.[[ExecutionContext]] to _context_.
-				1. Perform SetRealmGlobalObject(_realmRec_, *undefined*, *undefined*).
-				1. Perform ? SetDefaultGlobalBindings(_O_.[[ShadowRealm]]).
-				1. Perform ? HostInitializeShadowRealm(_O_.[[ShadowRealm]]).
+				1. Let _realmRec_ be the Realm of _context_.
+				1. Set _O_.[[ShadowRealm]] to _realmRec_.
+				1. Perform ? HostInitializeShadowRealm(_realmRec_).
```
caridy, line 480: What is the execution context after this? It seems to me that the new `InitializeHostDefinedRealm` "Push _newContext_ onto the execution context stack; _newContext_ is now the running execution context.", which is not the intent of `new ShadowRealm()`. Either `InitializeHostDefinedRealm` needs some work, or we must restore the execution context after calling it, isn't it?

spec.html
```diff
@@ -472,16 +472,12 @@ <h1>ShadowRealm ( )</h1>
...
 				1. If NewTarget is *undefined*, throw a *TypeError* exception.
 				1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%ShadowRealm.prototype%"*, « [[ShadowRealm]], [[ExecutionContext]] »).
-				1. Let _realmRec_ be CreateRealm().
-				1. Set _O_.[[ShadowRealm]] to _realmRec_.
-				1. Let _context_ be a new execution context.
-				1. Set the Function of _context_ to *null*.
-				1. Set the Realm of _context_ to _realmRec_.
-				1. Set the ScriptOrModule of _context_ to *null*.
+				1. Perform ? InitializeHostDefinedRealm().
+				1. Let _context_ be the running execution context.
```
caridy, line 476: I wasn't aware of https://github.com/tc39/ecma262/pull/3139, but `InitializeHostDefinedRealm` doing the push of the context without restoring it seems like a very problematic pattern, and the name of the AO doesn't reveal that. I don't recall a prior art for such pattern either.

spec.html
Ms2ger: Good catch. I've changed this PR; I'm not planning to raise this with 262 myself, but can adjust here if someone else does.

caridy (approved):
Ok, I'm fine with this as it stands. I will raise the issue in 262, to see if we can correct course there.

PR #412, opened 2024-11-10
docs: Fix link to issues

regseb:
Fix link to issues in [README](https://github.com/tc39/proposal-shadowrealm/blob/9321d6f/README.md#history).

Replace 
> we moved on from the exposed globalThis model to a lean isolated realms API (see #&NoBreak;289 and #&NoBreak;291)

by

> we moved on from the exposed globalThis model to a lean isolated realms API (see https://github.com/tc39/proposal-shadowrealm/issues/289 and https://github.com/tc39/proposal-shadowrealm/issues/291)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ See some examples [in the Explainer file](explainer.md).
 
 ## <a name='History'></a>History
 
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Agreed, it's helpful to have links there! Thanks.

PR #413, opened 2024-11-15
Assert that the global object inside the Shadow Realm is ordinary.

Ms2ger:
@mhofman: please let me know if there's any requirements this doesn't capture.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -627,13 +627,13 @@ location: https://tc39.es/proposal-shadowrealm/
 					including `HTMLElement`, `localStorage`, `fetch`, etc.
 				</p>
 			</emu-note>
-			<emu-note type=editor>
-				<p>
-					The ShadowRealm constructor (<emu-xref href="#sec-shadowrealm"></emu-xref>)
-					creates a new global object as an ordinary object. This
-					means all properties from the global object are deletable.
-				</p>
-			</emu-note>
+		</emu-clause>
+		<emu-clause id="sec-shadowrealm-host-requirements">
+			<h1>Requirements on host-defined global objects</h1>
+			<p>
+				If a host defines that a specific global object is to be used for a ShadowRealm, that
+				object must be an ordinary object.
+			</p>
 		</emu-clause>
 	</emu-clause>
 </emu-clause>
--- a/spec.html
+++ b/spec.html
@@ -628,7 +628,7 @@ location: https://tc39.es/proposal-shadowrealm/
 			<h1>Requirements on host-defined global objects</h1>
 			<p>
 				If a host defines that a specific global object is to be used for a ShadowRealm, that
-				object must be an ordinary object, and any properties on it must be configurable.
+				object must be an extensible ordinary object, and any properties on it must be configurable.
 			</p>
 		</emu-clause>
 	</emu-clause>
--- a/spec.html
+++ b/spec.html
@@ -497,6 +497,7 @@ location: https://tc39.es/proposal-shadowrealm/
 				1. Let _realmRec_ be the Realm of _innerContext_.
 				1. Set _O_.[[ShadowRealm]] to _realmRec_.
 				1. Perform ? HostInitializeShadowRealm(_realmRec_, _innerContext_, _O_).
+				1. Assert: _realmRec_.[[GlobalObject]] is an ordinary object.
 				1. Return _O_.
 			</emu-alg>
 		</emu-clause>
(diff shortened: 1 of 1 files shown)


mhofman:
If this is something we can actually assert, that would work for me. In this case we should likely also assert that the object is extensible and that all own properties are configurable.

I suppose the [editor note for `HostInitializeShadowRealm`](https://github.com/tc39/proposal-shadowrealm/blob/c321cb3b92d194f76b5a7040f550b1787c537d42/spec.html#L617-L618) also needs updating.

nicolo-ribaudo:
I don't think an assertion is the right tool here. Assertions are just editorial notes that explicitly state properties are an algorithm that are already true due to the algorithm itself. They are purely informative, and removing all assertions from the spec should not change its meaning.

Instead, we should pass a boolean _requireOrdinaryGlobal_ to [InitializeHostDefinedRealm](https://tc39.es/ecma262/#sec-initializehostdefinedrealm)(), and prefix the checks in steps 12 and 14 with "If _requireOrdinaryGlobal_ is false, and ..."

nicolo-ribaudo:
Actually, looking at the HTML integration this would be easier for layering:
```
12. If the host requires use of a custom object to serve as _realm_'s global object, then
    a. If _requireOrdinaryGlobal_ is **false**, let _global_ be an object created for such purpose in a host-defined manner.
    b. Else, let _global_ be an ordinary object created for such purpose in a host-defined manner.
14. Else,
   c. Let _global_ be OrdinaryObjectCreate(_realm_.[[Intrinsics]].[[%Object.prototype%]]).
```

Also, note that the current proposed HTML integration defines the ShadowRealm global to behave _exactly_ like an ordinary object, but it's not an ordinary object. Specifically, the [[SetPrototypeOf]] implementation is exotic and defined at https://webidl.spec.whatwg.org/#platform-object-setprototypeof. It however happens to behave exactly as the ordinary one, because _is global prototype chain mutable_ returns true.

mhofman:
> the current proposed HTML integration defines the ShadowRealm global to behave _exactly_ like an ordinary object, but it's not an ordinary object.

As I mentioned in #411, that's fine:

> To clarify, while the host must not be allowed to observably use an exotic object, that requirement does not prevent an implementation to actually use an exotic object as global object, as long as the exotic behavior is observably indistinguishable from a regular object behavior.

The spec deals in observable behaviors. An implementation is always free to deviate from the letter of the spec as long as JS programs can never observe the deviation.



> Assertions are just editorial notes that explicitly state properties are an algorithm that are already true due to the algorithm itself. They are purely informative, and removing all assertions from the spec should not change its meaning.

How do you recommend we constrain hosts on only adding configurable properties to the global object? In general how do we enforce invariants on host hooks?

nicolo-ribaudo:
> How do you recommend we constrain hosts on only adding configurable properties to the global object? In general how do we enforce invariants on host hooks?

By not putting the "Assert:" annotation in front of it :stuck_out_tongue: Normally for host hooks we say "the host must do X, Y and Z" rather than not stating requirements on what the host can do and _then_ assert that when we previously called into the host it did X, Y and Z.

Ms2ger:
> Also, note that the current proposed HTML integration defines the ShadowRealm global to behave _exactly_ like an ordinary object, but it's not an ordinary object. Specifically, the [[SetPrototypeOf]] implementation is exotic and defined at https://webidl.spec.whatwg.org/#platform-object-setprototypeof. It however happens to behave exactly as the ordinary one, because _is global prototype chain mutable_ returns true.

Fwiw, this is fixed in https://github.com/whatwg/webidl/pull/1437

nicolo-ribaudo (approved):
This works. Note that when opening the PR in the ECMA-262 repo, the new requirement should be added to `InitializeHostDefinedRealm` and not to `HostInitializeShadowRealm`.

spec.html
```diff
@@ -626,13 +623,13 @@ <h1>
...
-					creates a new global object as an ordinary object. This
-					means all properties from the global object are deletable.
-				</p>
-			</emu-note>
+		</emu-clause>
+		<emu-clause id="sec-shadowrealm-host-requirements">
+			<h1>Requirements on host-defined global objects</h1>
+			<p>
+				If a host defines that a specific global object is to be used for a ShadowRealm, that
+				object must be an ordinary object, and any properties on it must be configurable.
```
mhofman, line 631: The global object must also be extensible (not sure if I explicitly stated that previously, but it kinda went without saying)

```suggestion
				object must be an extensible ordinary object, and any properties on it must be configurable.
```

mhofman (approved):
Besides one minor correction, LGTM

ptomato (approved):
Makes sense, LGTM with Mathieu's suggestion about "extensible"

Ms2ger:
Thanks!

PR #415, opened 2024-11-25
Adjust GetActiveScriptOrModule to shadow realms

Ms2ger:


spec.html
```diff
@@ -633,5 +636,36 @@ <h1>Requirements on host-defined global objects</h1>
...
+
+	<emu-clause id="sec-getactivescriptormodule" type="abstract operation">
+		<h1>GetActiveScriptOrModule ( ): a Script Record, a Module Record, or *null*</h1>
+		<dl class="header">
+			<dt>description</dt>
+			<dd>It is used to determine the running script or module, based on the running execution context.</dd>
+		</dl>
+
+		<emu-alg>
+			1. If the execution context stack is empty, return *null*.
```
shannonbooth, line 658: While this appears to work for HTML integration in the test cases that I have run it against so far, unfortunately I have just found that it breaks some of the test-262 / test-js test cases (previously all test-262 cases were passing).

One of the failing test cases is this test case: https://github.com/LadybirdBrowser/ladybird/blob/5deb2b4cad3755c7e9251f6b23e9b1089b029aa7/Libraries/LibJS/Tests/builtins/ShadowRealm/ShadowRealm.prototype.importValue.js#L22

Where I am now getting:
```
 FAIL  builtins/ShadowRealm/ShadowRealm.prototype.importValue.js
    ❌ Suite:  normal behavior
         Test:   basic functionality (failed):
                 ExpectationError: Expected target to be null got TypeError: Cannot find/open module: './external-module.mjs'
```

I will try and investigate a bit more to find out what the problem is. Will also need to cross-check this against https://github.com/tc39/ecma262/pull/3374

